PR #262, opened 2026-04-02
feat(benchmark): add generalized ASAP benchmark pipeline with ClickHouse and Elasticsearch support

STWMichael:


milindsrivastava1997 (commented):
@kavyabhat02 could we add elastic instructions in the readme
@STWMichael @benjamib112 Do we need separate asap and clickhouse sql files in `generate_queries.py` now?

PR #327, opened 2026-05-06
feat: Elastic DSL in asap-planner-rs

Kolleida:


milindsrivastava1997:
## Code Review

### `infer_time_field` heuristics are overly broad (`extract_info.rs`)

```rust
let looks_like_time_field = field == "@timestamp"
    || field.contains("time")
    || field.contains("timestamp")
    || bound_is_time_like;
```

`field.contains("time")` matches `response_time`, `uptime`, `idle_time`, etc. A query with `range: { response_time: { gte: "now-5m" } }` would be incorrectly inferred as the time field even though `bound_is_time_like` already handles the datetime-bound case. The substring checks and `bound_is_time_like` are partially redundant — consider replacing the open-ended `contains` with more precise patterns (e.g. `field.ends_with("_time") || field == "timestamp"`).

---

### `t_lookback` ignores the query's actual time range (`elastic_dsl.rs`)

```rust
let t_lookback = self.t_repeat; // Default to repetition delay
```

For a query with `"gte": "now-5m"` and `repetition_delay: 60`, the lookback is 300s but `t_lookback = 60`. The SQL path derives the lookback from the query's time window; using `t_repeat` here will produce too-aggressive cleanup and evict data that queries still need. The actual `[gte, lte]` bounds from the parsed predicates should drive this value.

---

### `streaming_engine` is stored but never used (`elastic_dsl.rs`)

```rust
#[allow(dead_code)]
streaming_engine: StreamingEngine,
```

In the PromQL/SQL planners, `StreamingEngine` affects which aggregation types are emitted. If the same distinction doesn't apply here, a comment explaining why would help. If it does apply, this is a bug.

---

### `spatial_output` and `rollup` are always identical (`elastic_dsl.rs`)

```rust
let spatial = KeyByLabelNames::new(group_fields);
(spatial.clone(), spatial) // rollup == spatial_output
```

The comment above says "all potentially available fields become rollup when they're not in the group by" but the code sets rollup equal to spatial_output, not to the complement. Comment and implementation contradict each other — one of them needs to change.

---

### `_config` parameters suggest incomplete extraction (`generator.rs`)

Both `build_elastic_streaming_yaml` and `build_elastic_inference_yaml` accept `_config: &ElasticDSLControllerConfig` but never use it. Either wire it in or drop the parameter.

---

### Minor

- `get_group_by_fields` for `GroupBySpec::Filters` deduplicates with `Vec::contains` (O(n)) while `collect_elastic_metadata_fields` in the same file uses `IndexSet` — worth making consistent.
- `ElasticMappingSchema::add_index` consumes `self` (builder pattern) but `ElasticIndexSchema::new` doesn't — minor API surface inconsistency.

---

### Test gaps worth filling

- Conflicting time fields across queries on the same index (error path in `generator.rs`)
- `repetition_delay % data_ingestion_interval != 0` validation error
- `infer_time_field` fallback when no `@timestamp`-like range field is present

Kolleida:
Most of the nontrivial logic issues (e.g. rollup and lookback window) should be addressed.

milindsrivastava1997 (commented):
---
  Issue 1 — src/output/mod.rs is orphaned

  asap-planner-rs/src/output/mod.rs declares three submodules (generator, sql_generator, elastic_generator) but none of those files exist inside output/, and lib.rs never declares pub mod output;. The
  compiler never touches this directory, so it doesn't break the build — but it looks like an abandoned mid-refactor stub. Either delete the file or complete the move before merging.

  ---
  Issue 2 — Time range extraction uses .first() instead of matching the time field

  let time_range = query_info
      .predicates
      .first()
      .and_then(|p| range_query_to_time_range(p, 0));

  If a Term predicate (e.g. a service/datacenter filter) appears before the range predicate in the query, this returns None and silently falls back to t_repeat as the lookback window. Should scan for the
  predicate whose field matches query_info.time_field:

  let time_range = query_info.predicates.iter()
      .find(|p| matches!(p, Predicate::Range { field, .. } if field == &query_info.time_field))
      .and_then(|p| range_query_to_time_range(p, 0));

  ---

asap-common/dependencies/rs/elastic_dsl_utilities/src/ast_parsing/extract_info.rs
```diff
@@ -205,6 +208,26 @@ fn map_term_to_json_value(term: &dsl::Term) -> Option<TermValue> {
...
+                || field.ends_with("_time")
+                || bound_is_time_like;
+
+            if looks_like_time_field {
+                return field.clone();
+            }
+        }
+    }
+
+    "@timestamp".to_string()
```
milindsrivastava1997: Is this something Elastic adds or is this something that we're putting in as a fallback?

Kolleida: We're just putting this in as a fallback. `@timestamp` is a convention and would have to be added to data entries on the application side.

milindsrivastava1997: I'm wondering if we should just explicitly fail for now, instead of adding this hardcoded fallback.

Kolleida: Yeah, you're right, in hindsight failing would probably be better. I think I wanted to allow queries that don't specify a time field explicitly to still parse correctly, but it probably makes more sense to keep the time field inferring logic/metadata out of the parsing code entirely.

milindsrivastava1997: Yeah, let's fail for now. If we're building the system to work on time-series queries, that making that choice explicit is better. Please add the logic for this. Should be okay to merge after that.

milindsrivastava1997 (approved):

PR #364, opened 2026-05-24
feat(sql-utilities,query-engine): add SQL ORDER BY and LIMIT support

### LINKED ISSUE #363: feat(sql-utilities, query-engine): add SQL ORDER BY and LIMIT support for grouped aggregate queries.
ASAPQuery currently rejects grouped aggregate queries that include `ORDER BY` or `LIMIT`, even when the aggregation itself is already supported and precomputed.

For example, the following query falls through with destination=none_unsupported:
`SELECT proto, srcip, dstip,
       quantile(0.99)(pkt_len) AS p99_pkt_len
FROM netflow_table
WHERE time BETWEEN DATEADD(s, -11, NOW()) AND DATEADD(s, -10, NOW())
GROUP BY proto, srcip, dstip
ORDER BY p99_pkt_len DESC
LIMIT 10
`
This blocks common analytics patterns such as sorting aggregate results by percentile/sum/count metrics.

**Success Criteria**

- Queries with supported grouped aggregations continue to match the existing precomputed aggregation path when ORDER BY and/or LIMIT are present.
- ORDER BY on the aggregate alias or aggregate expression is accepted.
- LIMIT <N> is accepted.
- Existing aggregation matching and execution behavior remains unchanged aside from applying ordering/limit to the result set.

akanksha-akkihal:
### Summary

Add parser and post-processing support for SQL `ORDER BY` and `LIMIT` on supported grouped aggregate queries.

Queries such as:

```sql id="9r7xpw"
SELECT g1, g2, agg(v) AS p99
FROM t
GROUP BY g1, g2
ORDER BY p99 DESC
LIMIT 10
```

now parse and execute successfully where they previously fell through to:

```text id="8j1mde"
destination=none_unsupported
```

### Changes

* Capture:

  * `ORDER BY` items
  * `LIMIT N`
  * aggregate aliases

during SQL parsing and store them on `SQLQueryData`.

* Apply sorting and truncation after `execute_query_pipeline()` returns.

* Add a new `SqlPostProcessing` side-channel returned from `build_sql_query_pieces()` alongside `QueryExecutionContext`:

  * keeps the shared execution context unchanged
  * avoids impact to PromQL and Elastic paths

* `matches_sql_pattern()` now ignores:

  * `ORDER BY`
  * `LIMIT`
  * aggregate aliases

so existing registered templates continue matching queries that add presentation-only clauses.

* `SqlPostProcessing::apply()` short-circuits when no ordering or limit is present, making the fast path effectively zero-cost.

### Validation Rules

`ORDER BY` items must reference either:

* the aggregate alias
* a `GROUP BY` column

Unknown identifiers are rejected at parse time to avoid silently returning unsorted results for typo'd columns.

### Supported

* single or multi-column `ORDER BY`
* mixed `ASC` / `DESC`
* ordering by aggregate alias
* ordering by `GROUP BY` columns
* `LIMIT N`

### Still Unsupported

The following remain rejected:

* positional `ORDER BY`
* expressions in `ORDER BY`
* `OFFSET`
* `LIMIT BY`
* MySQL `LIMIT a, b`
* `WITH FILL`
* `NULLS FIRST/LAST`
* ClickHouse `ORDER BY ALL`

### Tests Added

Added 16 tests covering:

* aggregate alias ordering with `LIMIT`
* default ascending order
* multi-column mixed directions
* ordering by group-by keys
* limit-only queries
* tie-breaking behavior
* NaN handling
* `matches_sql_pattern()` compatibility
* no-op fast path
* end-to-end `SqlPostProcessing::apply`
* rejection of:

  * unknown columns
  * expressions
  * `OFFSET`

### DIFF
--- a/asap-query-engine/src/engines/simple_engine/sql.rs
+++ b/asap-query-engine/src/engines/simple_engine/sql.rs
@@ -5,18 +5,137 @@
 use super::SimpleEngine;
 use super::{QueryExecutionContext, QueryMetadata, QueryTimestamps};
 use crate::data_model::{AggregationIdInfo, QueryConfig, SchemaConfig};
-use crate::engines::query_result::QueryResult;
+use crate::engines::query_result::{InstantVector, InstantVectorElement, QueryResult};
 use asap_types::query_requirements::QueryRequirements;
 use asap_types::utils::normalize_spatial_filter;
 use promql_utilities::data_model::KeyByLabelNames;
 use promql_utilities::query_logics::enums::{QueryPatternType, Statistic};
 use sql_utilities::ast_matching::QueryType;
 use sql_utilities::ast_matching::{SQLPatternMatcher, SQLPatternParser, SQLQuery};
-use sql_utilities::sqlhelper::{AggregationInfo, SQLQueryData};
+use sql_utilities::sqlhelper::{AggregationInfo, OrderByItem, SQLQueryData};
 use sqlparser::dialect::*;
 use sqlparser::parser::Parser as parser;
 use std::collections::HashMap;
 use tracing::{debug, warn};
+
+/// SQL-only post-processing produced alongside a `QueryExecutionContext`:
+/// rules for ordering and truncating the final result vector.
+///
+/// Lives outside `QueryExecutionContext` so that PromQL/Elastic engines —
+/// which share that context but have no SQL-level ORDER BY / LIMIT — never
+/// have to know about these fields.
+#[derive(Debug, Clone, Default)]
+pub struct SqlPostProcessing {
+    /// Alias of the aggregate function in SELECT, e.g. `agg(v) AS p99`.
+    /// Used so `ORDER BY p99` resolves to `element.value`.
+    pub aggregation_alias: Option<String>,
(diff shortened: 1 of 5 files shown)


asap-query-engine/src/engines/simple_engine/sql.rs
```diff
@@ -202,15 +321,32 @@ impl SimpleEngine {
...
         query: String,
         time: f64,
     ) -> Option<QueryExecutionContext> {
+        self.build_sql_query_pieces(query, time).map(|(ctx, _)| ctx)
+    }
+
+    /// Internal: parses + plans a SQL query and returns both the execution
+    /// context (shared with PromQL/Elastic engines) and the SQL-only
+    /// post-processing rules (ORDER BY / LIMIT / alias resolution).
+    fn build_sql_query_pieces(
```
milindsrivastava1997: Can we maybe rename this to `build_query_execution_context_sql_with_post_processing`? I know it's verbose but "pieces" is vague.

akanksha-akkihal: Yeah, that makes sense. Will rename it.

PR #379, opened 2026-05-26
perf(precompute): avoid cloning the accumulator at window close

zzylol:
## Summary

`merge_panes_for_window` evicts the oldest pane destructively — it's `remove`d
from the map and dropped — but then called `take_accumulator()`, which **clones**
the inner accumulator before the updater is dropped. For sketch accumulators that
clone is expensive (a serialize → deserialize round-trip), so it's pure waste
when the pane is being evicted.

This adds `AccumulatorUpdater::into_accumulator(self: Box<Self>)` that **moves**
the inner accumulator out (no clone) and uses it for the evicted (oldest) pane in
`merge_panes_for_window`. Sliding-window shared panes still
`snapshot_accumulator()` (clone), since a later overlapping window may still read
them. A default trait impl falls back to a clone for updaters that can't cheaply
move their inner accumulator out.

For tumbling windows (one pane per window), this removes **one accumulator clone
per window close** on the ingest hot path.

### DIFF
--- a/asap-query-engine/src/precompute_engine/accumulator_factory.rs
+++ b/asap-query-engine/src/precompute_engine/accumulator_factory.rs
@@ -6,11 +6,12 @@ use crate::precompute_operators::{
 };
 use asap_types::aggregation_config::AggregationConfig;
 
-/// Generate the two boilerplate clone-based `AccumulatorUpdater` methods
+/// Generate the boilerplate `AccumulatorUpdater` extraction methods
+/// (`take_accumulator`/`snapshot_accumulator` clone, `into_accumulator` moves)
 /// for updaters whose inner `acc` field implements `Clone + AggregateCore`.
 /// Not applicable to `IncreaseAccumulatorUpdater` (its `acc` is `Option<_>`
 /// with non-trivial `None` handling).
-macro_rules! impl_clone_accumulator_methods {
+macro_rules! impl_accumulator_methods {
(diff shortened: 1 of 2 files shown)


milindsrivastava1997 (commented):
```
Code Review

  Correctness:

  IncreaseAccumulatorUpdater doesn't use the macro (its acc is Option<_> with special None handling), so it falls through to the default into_accumulator:
  fn into_accumulator(self: Box<Self>) -> Box<dyn AggregateCore> {
      self.snapshot_accumulator()  // default: still clones!
  }
  But take_accumulator for this type already does a zero-copy move via self.acc.take(). It would be consistent (and correct, not just faster) to override into_accumulator here too:
  fn into_accumulator(self: Box<Self>) -> Box<dyn AggregateCore> {
      let this = *self;
      Box::new(this.acc.unwrap_or_else(|| {
          IncreaseAccumulator::new(Measurement::new(0.0), 0, Measurement::new(0.0), 0)
      }))
  }
  IncreaseAccumulator is just scalar fields so the clone is cheap, but the inconsistency is a latent footgun if someone copies the pattern.

  Naming:

  The macro is called impl_clone_accumulator_methods but now also generates into_accumulator which is explicitly a move, not a clone. The name is now misleading. Worth renaming to impl_accumulator_methods or similar.
```

milindsrivastava1997 (approved):

PR #385, opened 2026-05-31
feat(asap-planner): support SQL COUNT(DISTINCT) -> HLL streaming/inference config generation

### LINKED ISSUE #384: feat(asap-planner) : support SQL COUNT(DISTINCT) -> HLL streaming/inference config generation
The query engine and SQL parser support COUNT(DISTINCT col) (normalized to CARDINALITY) with `AggregationType::HLL ` precompute, but asap-planner fails when generating inference_config.yaml / streaming_config.yaml for these queries.

### Current State
`asap-planner-rs/src/planner/sql.rs ` — `get_sql_statistics()` only maps: `QUANTILE, SUM, COUNT, AVG, MIN, MAX`
 It does not map `CARDINALITY`.

### Gaps
- `map_statistic_to_precompute_operator()` — no `Statistic::Cardinality` → HLL
- `build_sketch_parameters()` — no HLL / precision parameter path
- `SketchParameterOverrides` — no HLL block in planner input YAML (optional)
- For `COUNT(DISTINCT dstip) GROUP BY srcip` , the distinct column must be the sketch’s value column, not a rollup label—so the planner should put only srcip in grouping and leave rollup empty, instead of treating every non – GROUP BY metadata field (dstip, proto, etc.) as rollup dimensions.

akanksha-akkihal:
# Support HLL-backed `CARDINALITY` Aggregations

## Summary

Add end-to-end support for SQL `CARDINALITY` using HyperLogLog (HLL) sketches, including planner mapping, label routing, and configurable HLL precision.

### `promql_utilities` — `query_logics/logics.rs`

* Map:

  ```rust
  Statistic::Cardinality
  ```

  to:

  ```rust
  (AggregationType::HLL, "")
  ```

* Set:

  ```rust
  AggregationType::HLL => false
  ```

  in `does_precompute_operator_support_subpopulations()` so:

  * `grouping` = SQL `GROUP BY` columns
  * `aggregated` = empty

  Distinct values are stored in `value_column`.

* Test: `test_cardinality_maps_to_hll`

### `asap-planner-rs` — `planner/sql.rs`

* Map:

  ```sql
  CARDINALITY(...)
  ```

  to:

  ```rust
  vec![Statistic::Cardinality]
  ```

* For cardinality queries:

  * `rollup = []`
  * `grouping = GROUP BY columns`

  The distinct column is sketched via `value_column`, not treated as a rollup dimension.

### `asap-planner-rs` — `planner/sketch.rs`

* Add `AggregationType::HLL` support using:

  ```rust
  parameters.precision
  ```

* Default precision:

  ```text
  14
  ```

### `asap-planner-rs` — `config/input.rs`

* Add:

  ```rust
  HllParams
  ```

* Support overrides:

  ```yaml
  sketch_parameters:
    HLL:
      precision: <value>
  ```

### `asap-planner-rs` — `tests/sql_integration.rs`

* Add `spatial_count_distinct_hll` validating:

  * 1s spatial window
  * HLL aggregation
  * `value_column: dstip`
  * `grouping: [srcip]`
  * empty `rollup`
  * no `DeltaSetAggregator`

## Example

```sql
SELECT srcip, CARDINALITY(dstip)
FROM netflow
GROUP BY srcip
```

Produces:

```text
aggregation: HLL
value_column: dstip
grouping: [srcip]
rollup: []
```

with one HLL sketch per `srcip` group.

### DIFF
--- a/asap-common/dependencies/rs/promql_utilities/src/query_logics/logics.rs
+++ b/asap-common/dependencies/rs/promql_utilities/src/query_logics/logics.rs
@@ -50,6 +50,7 @@ pub fn map_statistic_to_precompute_operator(
             Ok((AggregationType::MultipleIncrease, "".to_string()))
         }
         Statistic::Topk => Ok((AggregationType::CountMinSketchWithHeap, "topk".to_string())),
+        Statistic::Cardinality => Ok((AggregationType::HLL, "".to_string())),
         _ => Err(format!("Statistic {statistic:?} not supported")),
     }
 }
@@ -81,6 +82,8 @@ pub fn does_precompute_operator_support_subpopulations(
 
         // CountMinSketchWithHeap is only supported for Topk — does not support subpopulations
         AggregationType::CountMinSketchWithHeap if matches!(statistic, Statistic::Topk) => false,
+
+        AggregationType::HLL => false,
 
         // Default: not supported
         _ => panic!("Unexpected precompute operator: {}", precompute_operator),
@@ -167,6 +170,16 @@ mod tests {
             Statistic::Sum,
             AggregationType::CountMinSketch,
         ));
+    }
+
+    #[test]
+    fn test_cardinality_maps_to_hll() {
+        let result = map_statistic_to_precompute_operator(
+            Statistic::Cardinality,
+            QueryTreatmentType::Approximate,
+        )
+        .unwrap();
+        assert_eq!(result, (AggregationType::HLL, "".to_string()));
     }
 
     #[test]
--- a/asap-common/dependencies/rs/promql_utilities/src/query_logics/logics.rs
+++ b/asap-common/dependencies/rs/promql_utilities/src/query_logics/logics.rs
@@ -51,6 +51,7 @@ pub fn map_statistic_to_precompute_operator(
         }
         Statistic::Topk => Ok((AggregationType::CountMinSketchWithHeap, "topk".to_string())),
         Statistic::Cardinality => Ok((AggregationType::HLL, "".to_string())),
+        _ => Err(format!("Statistic {statistic:?} not supported")),
     }
 }
 
--- a/asap-common/dependencies/rs/promql_utilities/src/query_logics/logics.rs
+++ b/asap-common/dependencies/rs/promql_utilities/src/query_logics/logics.rs
@@ -51,6 +51,7 @@ pub fn map_statistic_to_precompute_operator(
         }
         Statistic::Topk => Ok((AggregationType::CountMinSketchWithHeap, "topk".to_string())),
         Statistic::Cardinality => Ok((AggregationType::HLL, "".to_string())),
+        #[allow(unreachable_patterns)]
         _ => Err(format!("Statistic {statistic:?} not supported")),
     }
 }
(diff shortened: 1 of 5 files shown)


asap-common/dependencies/rs/promql_utilities/src/query_logics/logics.rs
```diff
@@ -50,7 +50,7 @@ pub fn map_statistic_to_precompute_operator(
             Ok((AggregationType::MultipleIncrease, "".to_string()))
         }
         Statistic::Topk => Ok((AggregationType::CountMinSketchWithHeap, "topk".to_string())),
-        _ => Err(format!("Statistic {statistic:?} not supported")),
```
milindsrivastava1997, line 53: Shouldn't this error condition be preserved?

asap-planner-rs/src/planner/sql.rs
```diff
@@ -105,7 +105,12 @@ impl SQLSingleQueryProcessor {
 
         // Label routing
         let spatial_output = KeyByLabelNames::new(labels.iter().cloned().collect::<Vec<_>>());
-        let rollup = all_metadata.difference(&spatial_output);
+        let rollup = if agg_info.get_name().eq_ignore_ascii_case("CARDINALITY") {
```
milindsrivastava1997: Rollup decision should be derived from the typed `Statistic::Cardinality` rather than a string

asap-common/dependencies/rs/promql_utilities/src/query_logics/logics.rs
akanksha-akkihal: Yes, I missed it. Will keep the error condition.

akanksha-akkihal: Clippy was giving me an Error
 `error: unreachable pattern
  --> asap-common/dependencies/rs/promql_utilities/src/query_logics/logics.rs:54:9
   |
54 |         _ => Err(format!("Statistic {statistic:?} not supported")),
   |         ^ no value can reach this`
   
   So, I am adding ` #[allow(unreachable_patterns)]`

milindsrivastava1997: Oh if it's unreachable, then it's okay, you can remove the error condition.

asap-planner-rs/src/planner/sql.rs
akanksha-akkihal: Sure, I will change it.

asap-common/dependencies/rs/promql_utilities/src/query_logics/logics.rs
akanksha-akkihal: With `#[allow(unreachable_patterns)]` and the _ arm: if someone adds a new Statistic but forgets to update this match, the project still builds, and callers get a runtime `Err("Statistic … not supported")`.

If we remove the _ arm: adding a new variant without a matching arm causes a compile error, so we can’t merge until the mapping is updated—no silent runtime failure.

Which behavior do we want?

milindsrivastava1997: remove the _ arm

milindsrivastava1997 (approved):

PR #387, opened 2026-06-01
fix(asap-planner): use MinMax with grouping for 1s spatial SQL MIN/MAX instead of MultipleMinMax

### LINKED ISSUE #386: fix(asap-planner) : emit MinMax with GROUP BY grouping for spatial SQL MIN/MAX
## Problem

For 1-second spatial ClickHouse queries using `MIN` / `MAX` with `GROUP BY`  
(e.g., `MAX(pkt_len) GROUP BY dstip`), the `asap-planner` was generating the wrong configuration.

It was using **MultipleMinMax**, with:
- grouping: empty  
- aggregated: `GROUP BY` columns (like `dstip`)  
- rollup: all other columns (`srcip`, `proto`, etc.)

This setup treats all data as one big group with sub-groups inside it.

However, the query engine expects a simpler setup using **MinMax**:
- grouping: `GROUP BY` columns (`dstip`)  
- aggregated: empty  
- rollup: empty  

This means: compute one min/max per `dstip` every second.

Because of this mismatch:
- Queries failed or returned empty results (e.g., “no precomputed outputs”)
- Even when data ingestion was working correctly  
- Since the planner config didn’t match what the engine expects

**Note:**  
This issue only affected **1-second spatial queries**.  
Other queries (like multi-second or temporal queries) were already correct.

akanksha-akkihal:
#### Before (default behavior)
Spatial `MIN` / `MAX` queries used:
- **Aggregation:** `MultipleMinMax`  
- **grouping:** `[]`  
- **aggregated:** `[GROUP BY columns]`  
- **rollup:** `[other metadata columns]`  

---

#### After

- Introduced `let mut configs` to allow post-processing updates  

- **Spatial `MIN` / `MAX` fix**  
  (applies only to `QueryType::Spatial` + `"MIN"` / `"MAX"`):
  - `aggregation_type` → `MinMax`  
  - `aggregation_sub_type` → `"min"` or `"max"`  
  - `grouping_labels` → SQL `GROUP BY` columns (`spatial_output`)  
  - `aggregated_labels` → `[]`  
  - `rollup_labels` → `[]`  

---

#### Unchanged
- Temporal / spatiotemporal `MIN` / `MAX` still use `MultipleMinMax`  

---

#### New Test: `spatial_max_pkt_len`

- **Query:**  
  `MAX(pkt_len) ... GROUP BY dstip` (1-second window, `netflow_table`)

---

## Runtime Impact (Generated Config)

| Field       | Before              | After                     |
|------------|--------------------|---------------------------|
| Type       | MultipleMinMax     | MinMax                    |
| grouping   | []                 | [dstip] (GROUP BY cols)   |
| aggregated | [dstip]            | []                        |
| rollup     | e.g. [srcip, proto]| []                        |

---

## Result

- Aligns planner output with:
  - query engine expectations  
  - ingest sharding (group-by as store key)

### DIFF
--- a/asap-planner-rs/src/planner/sql.rs
+++ b/asap-planner-rs/src/planner/sql.rs
@@ -114,7 +114,7 @@ impl SQLSingleQueryProcessor {
             all_metadata.difference(&spatial_output)
         };
 
-        let configs = build_agg_configs_for_statistics(
+        let mut configs = build_agg_configs_for_statistics(
             &statistics,
             treatment_type,
             &spatial_output,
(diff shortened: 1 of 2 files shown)


akanksha-akkihal:
Not fixing. Requirements changed.

PR #389, opened 2026-06-04
feat(query-engine): SQL Top-K via CountMinSketchWithHeap

### LINKED ISSUE #388: feat(query-engine): SQL Top-K via CountMinSketchWithHeap (COUNT + ORDER BY + LIMIT)
# Support SQL Top-K via CountMinSketchWithHeap (COUNT + ORDER BY + LIMIT)

## Summary

Support SQL top-k heavy-hitter queries using `CountMinSketchWithHeap`, for example:

```sql
SELECT srcip, COUNT(pkt_len) AS transfer_events
FROM netflow_table
WHERE time BETWEEN DATEADD(s, -11, NOW()) AND DATEADD(s, -10, NOW())
GROUP BY srcip
ORDER BY transfer_events DESC
LIMIT 10
```

The execution path should mirror PromQL `topk(k, ...)`:

* Precompute with `CountMinSketchWithHeap`
* Retrieve candidate keys from the sketch heap at query time
* Estimate counts from the CMS
* Sort descending by estimated count
* Return the top-k results

Results are approximate and based on CMS + heap estimates rather than exact ClickHouse aggregation.

## Motivation

Current SQL support for `ORDER BY ... DESC LIMIT k` operates only as a post-processing step and requires materializing all group-by keys. This prevents use of the sketch heap and defeats the purpose of approximate top-k precomputation.

Supporting SQL top-k enables:

* Efficient heavy-hitter queries without full group enumeration
* Reuse of existing PromQL top-k infrastructure
* Correct COUNT semantics (count events, not sum values)

### No SQL top-k detection

`build_query_execution_context_sql` mapped `COUNT(...)` to `Statistic::Count`, never `Statistic::Topk`. As a result, queries containing `ORDER BY <count_alias> DESC LIMIT k` were treated as ordinary grouped aggregations followed by `SqlPostProcessing`. The engine had to materialize all group-by keys before sorting and truncating, preventing use of the sketch heap and defeating the purpose of approximate top-k precomputation.

### Wrong precompute accumulator

`accumulator_factory.rs` routed `CountMinSketchWithHeap` to `CmsAccumulatorUpdater`, which only understands a plain Count-Min Sketch and ignores the heap component entirely. While count estimates could still be queried for known keys, `get_keys()` returned `None`, leaving the execution layer with no way to retrieve heavy-hitter candidates from the sketch itself.

### COUNT vs SUM weighting

`CmsAccumulatorUpdater` used the incoming sample value as the CMS weight. This is appropriate for sum-like statistics but incorrect for event counts. For a query such as `COUNT(pkt_len)`, each observation should contribute a weight of `1`, regardless of the value of `pkt_len`. The existing behavior effectively ranked groups by accumulated packet length rather than by number of events.

### Single `enable_topk` flag

A single `enable_topk` flag controlled two unrelated behaviors: (a) heap-based candidate enumeration and top-k truncation, and (b) PromQL-specific `__name__` label prefixing. SQL top-k requires the former but not the latter. Coupling both behaviors behind the same flag made it impossible to reuse the existing top-k machinery cleanly for SQL queries.

### Capability fallback vs heap-only design

The SQL top-k design uses the GROUP BY column as the sketch's aggregated dimension, resulting in empty `grouping_labels` and one sketch per window. However, capability matching relies on strict label-layout equality and treats `CountMinSketchWithHeap` as a multi-population value type. Without an explicit query configuration, fallback matching still expected a paired key aggregation (e.g. `SetAggregator` / `DeltaSetAggregator`) and therefore could not resolve to the intended heap-only execution path.

## Success Criteria

* SQL queries of the form:

```sql
SELECT g, COUNT(v)
FROM t
GROUP BY g
ORDER BY COUNT(v) DESC
LIMIT k
```

are recognized as top-k queries.

* Planner generates `CountMinSketchWithHeap` precomputes with the requested `k`.

* Query execution retrieves candidate keys from the heap rather than enumerating all groups.

* COUNT uses weight `1` per event when updating the sketch.

* Results are sorted by estimated count and truncated to `k`.

* Existing PromQL top-k behavior remains unchanged.

* Non-top-k SQL aggregations continue to use existing execution paths.

akanksha-akkihal:
## Overview

Adds an end-to-end SQL execution path for approximate top-k queries by event count, aligned with the existing PromQL `topk(k, ...)` implementation. Queries of the form:

```sql
SELECT srcip, COUNT(pkt_len) AS transfer_events
FROM netflow_table
WHERE time BETWEEN DATEADD(s, -11, NOW()) AND DATEADD(s, -10, NOW())
GROUP BY srcip
ORDER BY transfer_events DESC
LIMIT 10
```

are now recognized as top-k queries during SQL planning, mapped to `Statistic::Topk`, and executed using `CountMinSketchWithHeap`. Instead of materializing all group-by keys and applying `ORDER BY` / `LIMIT` as a post-processing step, the engine retrieves candidate keys directly from the sketch heap, estimates counts from the CMS, sorts them, and returns the top-k results.

### Precompute (`accumulator_factory.rs`)

* Introduced `CmsWithHeapAccumulatorUpdater` for `CountMinSketchWithHeap`.
* Replaces the previous routing to `CmsAccumulatorUpdater`, which treated the sketch as a plain CMS and could not enumerate heap candidates.
* Added `count_events` support (default `true`), allowing COUNT-style workloads to update the sketch with a constant weight of `1.0` per observation rather than using the sample value as the weight.
* Added `cms_heap_params()` to construct heap-backed sketches from planner-generated parameters.
* Supports both planner/Arroyo naming (`depth`, `width`, `heapsize`) and legacy naming (`row_num`, `col_num`, `heap_size`), with defaults of `3`, `1024`, and `32`.

### SQL Engine (`sql.rs`)

* Added `detect_sql_topk()` to identify SQL top-k patterns and promote them to `Statistic::Topk`.
* Detection requires:

  * `COUNT(...)`
  * `GROUP BY`
  * `ORDER BY <aggregate_alias> DESC`
  * `LIMIT k`
* Extracts and stores `k` in `query_kwargs` for downstream execution.
* Context generation uses `Statistic::Topk` and empty `grouping_labels`, matching the intended sketch layout where the GROUP BY column is represented as the sketch's aggregated dimension rather than a storage partition key.
* Updated `handle_query_sql()` so top-k queries execute through the sketch heap path with:

  * `enable_topk_limiting = true`
  * `enable_topk_formatting = false`
* SQL top-k queries bypass `SqlPostProcessing`, since ordering and truncation are now performed directly by the query pipeline.
* Added unit tests covering:

  * Top-k query detection
  * Positive and negative detection cases
  * End-to-end execution against a seeded sketch

### Query Pipeline (`mod.rs`)

* Split the previous `enable_topk` flag into two independent controls:

  * `enable_topk_limiting`
  * `enable_topk_formatting`
* `enable_topk_limiting` sorts top-k candidates and truncates results to `k`.
* `enable_topk_formatting` preserves the existing PromQL behavior of prepending metric names to labels.
* Heap candidate collection no longer truncates during enumeration.
* The full candidate set is collected first, then sorted and truncated within the pipeline, ensuring consistent ordering behavior across execution paths.

### Call Sites

* **PromQL**

  * `enable_topk_limiting = true`
  * `enable_topk_formatting = true`
  * Existing behavior unchanged.

* **SQL Top-K**

  * `enable_topk_limiting = true`
  * `enable_topk_formatting = false`

* **SQL (non-top-k) / Elastic**

  * `enable_topk_limiting = false`
  * `enable_topk_formatting = false`
  * Existing execution and post-processing paths unchanged.

### Tests and Supporting Changes

* Updated `engine_factories.rs` to accommodate the new `execute_query_pipeline` signature.
* Added SQL top-k detection tests covering supported and unsupported query shapes.
* Added pipeline tests validating context construction and heap-based execution using pre-seeded sketches.
* Preserved existing PromQL top-k behavior while validating SQL-specific execution semantics.

### DIFF
--- a/asap-query-engine/src/engines/simple_engine/sql.rs
+++ b/asap-query-engine/src/engines/simple_engine/sql.rs
@@ -386,7 +386,7 @@ impl SimpleEngine {
         // metric-name formatting; the sketch heap already produces the ranked
         // `(group-by key, count)` rows, so SQL ORDER BY / LIMIT post-processing
         // would be redundant and is skipped.
-        let (output_labels, result) = self.execute_context(context, is_topk, false)?;
+        let (output_labels, result) = self.execute_context(context, is_topk, false, false)?;
         let result = if is_topk {
             result
         } else {
--- a/asap-query-engine/src/engines/simple_engine/sql.rs
+++ b/asap-query-engine/src/engines/simple_engine/sql.rs
@@ -137,6 +137,50 @@ fn sort_and_truncate_instant_vector(
     results
 }
 
+/// Detect a SQL top-k query and return its `k`.
+///
+/// Recognises the heavy-hitter shape that `CountMinSketchWithHeap` serves:
+///
+/// ```sql
+/// SELECT <key>, COUNT(<col>) AS <alias>
+/// FROM <table> WHERE <1s window>
+/// GROUP BY <key>
+/// ORDER BY <alias> DESC
+/// LIMIT k
+/// ```
+///
+/// The grouping key (`<key>`) becomes the *aggregated* dimension inside the
+/// sketch's heap — not a precompute partition key — so a single sketch per
+/// window tracks the top keys by event count.
+///
+/// The SQL parser only accepts identifier ORDER BY targets, so the descending
+/// order must reference the aggregate's alias (e.g. `transfer_events`), not the
+/// `COUNT(col)` expression itself.
+pub(crate) fn detect_sql_topk(query_data: &SQLQueryData) -> Option<u64> {
+    let k = query_data.limit?;
+    // Need a GROUP BY key to rank and an ORDER BY to define "top".
+    if query_data.labels.is_empty() || query_data.order_by.is_empty() {
+        return None;
+    }
+    // CountMinSketchWithHeap tracks heavy hitters by COUNT.
+    if !query_data
+        .aggregation_info
+        .get_name()
+        .eq_ignore_ascii_case("count")
+    {
+        return None;
+    }
+    // Primary ordering must be the aggregate alias, descending (largest first).
+    let primary = &query_data.order_by[0];
+    if primary.ascending {
+        return None;
+    }
+    if query_data.aggregation_alias.as_deref() != Some(primary.column.as_str()) {
+        return None;
+    }
+    Some(k)
+}
+
 impl SimpleEngine {
     /// Finds the query configuration for a SQL query using structural pattern matching.
     ///
@@ -268,6 +312,7 @@ impl SimpleEngine {
         &self,
         match_result: &SQLQuery,
         query_pattern_type: QueryPatternType,
+        topk_k: Option<u64>,
     ) -> QueryRequirements {
         let query_data = match_result
             .outer_data()
@@ -284,9 +329,18 @@ impl SimpleEngine {
             _ => query_data.aggregation_info.get_name().to_lowercase(),
         };
 
-        let statistics: Vec<Statistic> = Self::parse_single_statistic(&statistic_name)
-            .into_iter()
-            .collect();
+        // For top-k the requirement is `Statistic::Topk` (→ CountMinSketchWithHeap)
+        // and the grouping is empty: the GROUP BY column is the sketch's
+        // *aggregated* (heavy-hitter) dimension, held inside one sketch per
+        // window, not a precompute partition key.
+        let is_topk = topk_k.is_some();
+        let statistics: Vec<Statistic> = if is_topk {
+            vec![Statistic::Topk]
+        } else {
+            Self::parse_single_statistic(&statistic_name)
+                .into_iter()
+                .collect()
+        };
 
         let data_range_ms = match query_pattern_type {
             QueryPatternType::OnlySpatial => None,
@@ -305,7 +359,11 @@ impl SimpleEngine {
             }
         };
 
-        let grouping_labels = KeyByLabelNames::new(query_data.labels.clone().into_iter().collect());
+        let grouping_labels = if is_topk {
+            KeyByLabelNames::empty()
+        } else {
+            KeyByLabelNames::new(query_data.labels.clone().into_iter().collect())
+        };
 
         QueryRequirements {
             metric,
@@ -323,8 +381,17 @@ impl SimpleEngine {
     ) -> Option<(KeyByLabelNames, QueryResult)> {
         let (context, post) =
             self.build_query_execution_context_sql_with_post_processing(query, time)?;
-        let (output_labels, result) = self.execute_context(context, false, false)?;
-        let result = post.apply(&output_labels, result);
+        let is_topk = context.metadata.statistic_to_compute == Statistic::Topk;
+        // Top-k: enable heap-based limiting (truncate to k) but NOT PromQL-style
+        // metric-name formatting; the sketch heap already produces the ranked
+        // `(group-by key, count)` rows, so SQL ORDER BY / LIMIT post-processing
+        // would be redundant and is skipped.
+        let (output_labels, result) = self.execute_context(context, is_topk, false)?;
+        let result = if is_topk {
+            result
+        } else {
+            post.apply(&output_labels, result)
+        };
         Some((output_labels, result))
     }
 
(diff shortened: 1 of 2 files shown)


asap-query-engine/src/engines/simple_engine/mod.rs
```diff
@@ -841,22 +857,29 @@ impl SimpleEngine {
...
-        let sorted_results: Vec<(Option<KeyByLabelValues>, f64)> =
-            if *statistic == Statistic::Topk && enable_topk_formatting {
-                // Sort by value descending for topk
-                let mut sorted: Vec<_> = unformatted_results.into_iter().collect();
-                sorted.sort_by(|a, b| b.1.partial_cmp(&a.1).unwrap_or(std::cmp::Ordering::Equal));
+        let sorted_results: Vec<(Option<KeyByLabelValues>, f64)> = if *statistic == Statistic::Topk
+        {
+            // Sort by value descending for topk (independent of output formatting).
+            let mut sorted: Vec<_> = unformatted_results.into_iter().collect();
+            sorted.sort_by(|a, b| b.1.partial_cmp(&a.1).unwrap_or(std::cmp::Ordering::Equal));
```
milindsrivastava1997, line 879: Why are we doing this independent of whether `enable_topk_formatting` is true?

asap-query-engine/src/engines/simple_engine/mod.rs
```diff
@@ -990,10 +1016,11 @@ impl SimpleEngine {
     fn execute_context(
         &self,
         context: QueryExecutionContext,
-        enable_topk: bool,
+        enable_topk_limiting: bool,
+        enable_topk_formatting: bool,
```
milindsrivastava1997, line 1020: Didn't understand the change in behavior from the single `enable_topk` flag earlier to the 2 separate flags right now

asap-query-engine/src/engines/simple_engine/mod.rs
```diff
@@ -1238,28 +1265,6 @@ impl SimpleEngine {
         Ok(unformatted_results)
     }
 
-    /// Limits keys for topk queries
-    fn limit_keys_for_topk(
-        &self,
-        keys: Vec<KeyByLabelValues>,
```
milindsrivastava1997: Why is this no longer needed?

asap-query-engine/src/engines/simple_engine/sql.rs
```diff
@@ -709,6 +790,111 @@ impl SimpleEngine {
...
+            "LIMIT without ORDER BY is not top-k"
+        );
+    }
+
+    #[test]
+    fn sum_aggregate_is_not_topk() {
+        let sql = format!(
+            "SELECT srcip, SUM(pkt_len) AS total FROM netflow_table {WINDOW} \
+             GROUP BY srcip ORDER BY total DESC LIMIT 10"
+        );
```
milindsrivastava1997, line 877: We should be able to support this query too

asap-query-engine/src/engines/simple_engine/mod.rs
akanksha-akkihal: Sorting is independent of `enable_topk_formatting` because ranking and label shape are two different concerns that SQL and PromQL need in different combinations.

The sketch heap is not value-sorted. We deliberately collect all heap candidates in `collect_all_results` and defer sort + truncate to the pipeline (the heap can hold `heapsize > k` entries and isn’t ordered by estimate). Sorting is part of top-k semantics, not PromQL output formatting.

`enable_topk_formatting` only controls the PromQL label prefix (prepend metric name to match {__name__, …} series shape). SQL leaves it `false` so rows stay bare `(srcip, count)`.

If sorting were behind `enable_topk_formatting`, SQL top-k would break: you’d get 10 rows in random order instead of the 10 highest counts.

asap-query-engine/src/engines/simple_engine/mod.rs
akanksha-akkihal: `limit_keys_for_topk` was removed because early truncation at collection time was incorrect for sketch-based top-k, and its job is now handled after we have counts and can sort.

When `enable_topk` was on, we took the first k keys from `get_keys()` before querying their values, then sorted/truncated later in the pipeline.

Why that’s wrong

- Heap key order ≠ rank order. get_keys() returns candidates in heap iteration order, not sorted by estimated count. take(k) there could drop a true top-k key that happens to appear after the first k in that order.

- heapsize is often > k. The sketch may hold e.g. 40 candidates for LIMIT 10. Pre-truncating to 10 at collection means we never even fetch estimates for the other 30 — some of which might be the actual top 10 once sorted by value.

- Limiting belongs after sorting. Top-k means “the k keys with the highest counts.” You need every heap candidate’s estimate, sort descending, then truncate — not truncate the key list first and hope heap order was good enough.

`enable_topk_limiting` is at lines `684 - 695 `

asap-query-engine/src/engines/simple_engine/sql.rs
akanksha-akkihal: This PR intentionally scoped SQL top-k detection to COUNT only, because the target netflow query was COUNT(pkt_len).
I’ll open a follow-up PR for SUM top-k so this one stays focused on the COUNT path.
Should I keep this test as of now or remove it?

milindsrivastava1997: Sounds good, keep it

milindsrivastava1997 (approved):

PR #391, opened 2026-06-05
refactor(query-engine): split enable_topk into limiting vs formatting flags

### LINKED ISSUE #390: refactor(query-engine): Split enable_topk into limiting vs formatting flags
## Summary

Refactor the query engine's single `enable_topk` boolean into two flags so PromQL and SQL top-k can share the heap-based ranking path without forcing PromQL-specific output shaping on SQL.

---

## Background

`execute_query_pipeline` / `execute_context` previously took one flag:
`execute_query_pipeline(context, enable_topk: bool)`

When `enable_topk` was `true`, the pipeline did all of the following together:

- **Limiting** — use sketch-heap candidates and truncate results to `k`
- **Sorting** — order rows by value descending (top-k semantics)
- **Formatting** — prepend the metric name to each row's labels (PromQL instant-vector / `{__name__, …}` shape)

That coupling worked for PromQL `topk(k, …)`, which needs all three.

It does **not** work for SQL top-k, which needs limiting + sorting but must **not** prepend the metric/table name to labels. SQL should return bare `(group-by columns, value)` rows, e.g. `(srcip, count)`.

With a single flag, SQL had no way to get heap-driven top-k without also getting PromQL label formatting.

---

## Proposed Change

Replace `enable_topk` with two separate flags:

| Flag | Purpose | PromQL top-k | SQL top-k | Other queries |
|---|---|---|---|---|
| `enable_topk_limiting` | Truncate sorted results to `k` | `true` | `true` (when top-k detected) | `false` |
| `enable_topk_formatting` | Prepend metric name to label values | `true` | `false` | `false` |

> **Note:** Sorting for `Statistic::Topk` remains always on (not gated by either flag). Both SQL and PromQL need descending order before `truncate(k)`; collection returns heap keys in a `HashMap` with undefined iteration order.

akanksha-akkihal:
### Background

Previously, `enable_topk` was responsible for two unrelated behaviors:

- **Limiting** → using the sketch heap to select top-k results (sort + truncate)
- **Formatting** → prepending the metric name (`__name__`) in PromQL-style output

This coupling worked for PromQL (which requires both behaviors), but it breaks SQL semantics.

SQL only needs **top-k selection**, not PromQL-style label formatting. With a single flag, it was impossible to express:

> “Use heap-driven top-k selection without applying PromQL metric name formatting.”

---

## Problem

The original design conflated two independent concerns:

- How results are **selected** (ranking / top-k computation)
- How results are **represented** (PromQL vs SQL output shape)

This caused SQL queries to unintentionally inherit PromQL formatting.

---

## Solution: Split into two flags

We introduce two independent flags:

### 1. `enable_topk_limiting`
Controls heap-based top-k selection:

- Uses sketch heap to identify top-k elements
- Applies sorting + truncation logic

### 2. `enable_topk_formatting`
Controls PromQL-specific output formatting:

- Prepends metric name (`__name__`)
- Produces PromQL instant-vector style output

---

## Resulting behavior

| Query Type | `enable_topk_limiting` | `enable_topk_formatting` | Behavior |
|------------|------------------------|----------------------------|----------|
| PromQL     | true                   | true                       | Top-k + metric prefix formatting |
| SQL top-k  | true                   | false                      | Top-k only (clean rows) |
| SQL/Elastic (non-top-k) | false | false | Standard execution |

---

## Subtle detail

Sorting remains enabled for `Statistic::Topk` regardless of formatting.

- Ensures correct descending order before truncation
- Prevents PromQL formatting logic from influencing SQL result structure

---

## TL;DR

This change decouples:

- **How we select top-k rows** (ranking logic)
from
- **How results are formatted** (PromQL vs SQL representation)

This separation aligns the execution model with the fundamental difference between SQL and PromQL semantics.

### DIFF
--- a/asap-query-engine/src/engines/simple_engine/elastic.rs
+++ b/asap-query-engine/src/engines/simple_engine/elastic.rs
@@ -25,7 +25,7 @@ impl SimpleEngine {
             "Built execution context for ElasticSearch query {:?}",
             context
         );
-        self.execute_context(context, false)
+        self.execute_context(context, false, false)
     }
 
     pub fn build_query_execution_context_elastic(
--- a/asap-query-engine/src/engines/simple_engine/mod.rs
+++ b/asap-query-engine/src/engines/simple_engine/mod.rs
@@ -626,27 +626,32 @@ impl SimpleEngine {
         merged_keys: Option<&HashMap<Option<KeyByLabelValues>, Box<dyn AggregateCore>>>,
         statistic: &Statistic,
         query_kwargs: &HashMap<String, String>,
-        enable_topk_limiting: bool,
     ) -> Result<HashMap<Option<KeyByLabelValues>, f64>, String> {
         if let Some(keys_map) = merged_keys {
             // Separate keys and values
             self.collect_results_separate_keys(merged_values, keys_map, statistic, query_kwargs)
         } else {
             // Same aggregation for keys and values
-            self.collect_results_same_aggregation(
-                merged_values,
-                statistic,
-                query_kwargs,
-                enable_topk_limiting,
-            )
+            self.collect_results_same_aggregation(merged_values, statistic, query_kwargs)
         }
     }
 
-    /// Executes the complete query pipeline: plan, execute, collect, and format
+    /// Executes the complete query pipeline: plan, execute, collect, and format.
+    ///
+    /// The two top-k flags are deliberately separate because the two engines
+    /// need different halves of the behaviour:
+    ///   * `enable_topk_limiting` — enumerate candidate keys from the sketch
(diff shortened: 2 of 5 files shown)


milindsrivastava1997 (commented):
Can you check if there are tests that exercise PromQL's `topk` behavior? both in terms of the semantics, as well as the extra output label that is inserted into the result? If those tests are not there, could you please add them (use main as a source for adding them), and then test them against this branch? 
Apart from that, looks good

akanksha-akkihal:
Currently, there are no tests that exercise PromQL's topk behavior. Shall I add the test to the same PR?

milindsrivastava1997 (approved):

PR #393, opened 2026-06-08
feat(query-engine): Support for SUM Top-k queries using CountMinSketchWithHeap

### LINKED ISSUE #392: feat(query-engine) : Support SQL SUM … ORDER BY … LIMIT top-k via CountMinSketchWithHeap
## Summary

Extend SQL top-k support so `SUM(col) … GROUP BY … ORDER BY <alias> DESC LIMIT k` is detected and served by `CountMinSketchWithHeap`, using the same resolution paths as existing `COUNT(col)` top-k queries. Add capability-matching logic to distinguish count-weighted vs sum-weighted heap configs via count_events.

## Background

SQL top-k is already supported for:

```sql
SELECT srcip, COUNT(pkt_len) AS transfer_events
FROM netflow_table
WHERE time BETWEEN … AND …
GROUP BY srcip
ORDER BY transfer_events DESC
LIMIT 10
```

This path:

* Detects the pattern via `detect_sql_topk`
* Promotes it to `Statistic::Topk`
* Routes to `CountMinSketchWithHeap` with `count_events: true` (unit weight per event)
* Resolves self-keyed when the inference query_config lists a single aggregation, or via the multi-population fallback (heap + SetAggregator/DeltaSetAggregator) when no query config is present

`SUM + ORDER BY + LIMIT` was not supported: detection rejected non-COUNT aggregates, and capability matching could not tell apart two heap configs on the same metric (count vs sum weighting).

## Problem

* **Detection gap:** `detect_sql_topk` only accepted `COUNT`, so `SUM(pkt_len) … ORDER BY total DESC LIMIT k` fell through to the generic SQL path (post-processing sort/limit on a non-sketch aggregation).
* **Capability-matching gap:** Two `CountMinSketchWithHeap` configs on the same metric (one `count_events: true`, one `count_events: false`) could not be disambiguated when falling back to capability matching.
* **Parity:** SUM top-k should follow the same resolution model as COUNT top-k (query-config self-keyed vs fallback separate-key), not introduce a new keying model.

## Acceptance criteria

* SUM(col) AS alias … GROUP BY key … ORDER BY alias DESC LIMIT k is detected as top-k and promoted to Statistic::Topk.

* COUNT top-k behavior is unchanged.

* Self-keyed SUM top-k works via a single-aggregation query_config (same as COUNT).

* Capability fallback selects the correct heap when both count- and sum-weighted configs exist on the same metric.

* SUM top-k does not match a count-only sketch (count_events: true).

* Unit tests pass: cargo test -p asap_types --lib capability_matching, cargo test -p query_engine_rust --lib topk.

akanksha-akkihal:
### Summary

Extends SQL top-k support so `SUM(col) … GROUP BY … ORDER BY <alias> DESC LIMIT k` is detected and served by `CountMinSketchWithHeap`, using the same resolution model as existing `COUNT(col)` top-k queries.

### `asap-query-engine/src/engines/simple_engine/sql.rs`

Main implementation file.

- **New types:** `TopkWeighting` (`Count` | `Sum`) and `SqlTopk { k, weighting }` with `count_events()` helper.
- **`detect_sql_topk`:** return type changed from `Option<u64>` → `Option<SqlTopk>`. Now accepts both `COUNT(col)` and `SUM(col)` (same ORDER BY alias DESC + LIMIT rules). Rejects other aggregates. Documents non-negative SUM assumption.
- **Context building:** `topk_k` replaced with `topk`; still promotes to `Statistic::Topk` and threads `k` into `query_kwargs`.
- **`build_query_requirements_sql`:** signature takes `Option<SqlTopk>` instead of `Option<u64>`; sets `topk_count_events` from detected weighting for capability fallback.
- **`detect_topk_tests`:**
  - COUNT test now asserts `TopkWeighting::Count`.
  - New `sum_order_by_alias_desc_limit_is_topk`.
  - `sum_aggregate_is_not_topk` replaced with `min_aggregate_is_not_topk`.
- **`topk_pipeline_tests`:**
  - New `sum_topk_query()` helper.
  - New `build_sum_topk_engine()` — single-aggregation query config, `count_events: false`.
  - New `sum_topk_resolves_self_keyed_heap` — mirrors COUNT self-keyed path (`key id == value id`, no `keys_query`).

### `asap-common/dependencies/rs/asap_types/src/query_requirements.rs` (+9 lines)

- **New field:** `topk_count_events: Option<bool>` on `QueryRequirements`.
  - `Some(true)` → COUNT top-k sketch (`count_events: true`)
  - `Some(false)` → SUM top-k sketch (`count_events: false`)
  - `None` → no weighting constraint

### `asap-common/dependencies/rs/asap_types/src/capability_matching.rs` 

Capability-matching fallback for count vs sum top-k.

- **New helpers:**
  - `config_count_events()` — reads `count_events` from heap config (default `true`).
  - `topk_weighting_compatible()` — filters `Statistic::Topk` candidates by requested weighting.
- **`find_compatible_aggregation`:** adds `topk_weighting_compatible(...)` to the candidate filter. Top-k still uses the standard multi-population path (heap + key agg in fallback; self-keyed via query config).
- **Test helper `req()`:** initializes `topk_count_events: None`.
- **New tests:** `topk_count_query_picks_count_events_sketch`, `topk_sum_query_picks_value_weighted_sketch`, `topk_sum_query_rejects_count_only_sketch`, `topk_count_query_matches_sketch_without_explicit_flag`, `topk_unconstrained_weighting_matches_any_sketch` (all provision a `DeltaSetAggregator` for the fallback path).

### `asap-query-engine/src/engines/simple_engine/promql.rs` 

- **`build_query_requirements_promql`:** sets `topk_count_events: None` so PromQL top-k does not constrain sketch weighting during capability matching.

### DIFF
--- a/asap-common/dependencies/rs/asap_types/src/capability_matching.rs
+++ b/asap-common/dependencies/rs/asap_types/src/capability_matching.rs
@@ -101,6 +101,39 @@ pub fn spatial_filter_compatible(config_filter: &str, req_filter: &str) -> bool 
     config_norm == req_norm
 }
 
+/// Reads the `count_events` parameter from a `CountMinSketchWithHeap` config.
+/// Defaults to `true` (COUNT semantics) so existing count top-k configs that
+/// omit the flag keep matching COUNT top-k queries.
+fn config_count_events(config: &AggregationConfig) -> bool {
+    config
+        .parameters
+        .get("count_events")
+        .and_then(|v| v.as_bool())
+        .unwrap_or(true)
+}
+
+/// Top-k weighting compatibility. Only constrains `Statistic::Topk` candidates;
+/// every other statistic passes unconditionally.
+///
+/// A COUNT top-k query (`Some(true)`) must be served by a `count_events: true`
+/// sketch and a SUM top-k query (`Some(false)`) by a `count_events: false`
+/// (value-weighted) sketch. This is what tells two `CountMinSketchWithHeap`
+/// configs on the same metric apart. `None` (non-top-k, or PromQL top-k which
+/// does not pin the weighting) imposes no constraint.
+pub fn topk_weighting_compatible(
+    stat: Statistic,
+    config: &AggregationConfig,
+    req_count_events: Option<bool>,
+) -> bool {
+    if stat != Statistic::Topk {
+        return true;
+    }
+    match req_count_events {
+        Some(want) => config_count_events(config) == want,
+        None => true,
+    }
+}
+
 /// Aggregation priority comparator: prefer larger `window_size` (descending).
 /// This is a separate function so callers can swap the policy without touching matching logic.
 pub fn aggregation_priority(a: &AggregationConfig, b: &AggregationConfig) -> Ordering {
@@ -157,7 +190,8 @@ pub fn find_compatible_aggregation(
                     && spatial_filter_compatible(
                         &c.spatial_filter_normalized,
                         &requirements.spatial_filter_normalized,
-                    );
+                    )
+                    && topk_weighting_compatible(stat, c, requirements.topk_count_events);
                 if !ok {
                     debug!(
                         agg_id = c.aggregation_id,
@@ -216,6 +250,10 @@ pub fn find_compatible_aggregation(
     }
 
     // If value type is multi-population, find the paired key aggregation.
+    // Top-k (CountMinSketchWithHeap) follows the same path as plain COUNT: the
+    // self-keyed case is expressed via the query_config path (a single
+    // aggregation reference), while the capability-matching fallback resolves a
+    // separate key aggregation just like any other multi-population value type.
     let key_agg: &AggregationConfig = if is_multi_population_value_type(value_agg.aggregation_type)
     {
         let ka = configs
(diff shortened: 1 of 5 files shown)


milindsrivastava1997 (commented):
1.  detect_sql_topk — topk detected from outer_data() only

  In build_query_execution_context_sql, detect_sql_topk is called with query_data from match_result.outer_data(). For a OneTemporalOneSpatial pattern the temporal aggregate lives in inner_data. SUM top-k only makes sense as a flat OnlyTemporal query, but there's no guard here — if a subquery-form SUM ORDER BY LIMIT accidentally matches, it would silently detect top-k from the outer data and produce confusing behavior. The COUNT path had the same latent issue before this PR, but it's worth noting since SUM makes the shape more likely to appear in nested queries.

2. Silent failure for negative SUM columns

  sql.rs:150-156 documents the non-negative assumption but there's no runtime check or even a warn!() at query time. A user querying SUM(delta) where the column can go negative will get plausible-looking but wrong top-k results silently. Even a single warn! when the SUM top-k path is taken ("SUM top-k assumes non-negative values; results are undefined for columns with negative entries") would be better than nothing.


3. Test gap: SUM capability-matching fallback path

  topk_pipeline_tests has sum_topk_resolves_self_keyed_heap, which exercises the query_config path (single-aggregation reference). There's no integration-level test for the capability-matching fallback (when no query_config matches the SUM query). The unit tests in capability_matching.rs cover the logic, but the wiring from the engine into the matcher is untested for SUM. The COUNT pipeline tests have the same gap. Also, the fact that count_events is true as a default is also not tested.

milindsrivastava1997 (approved):

PR #395, opened 2026-06-08
feat(asap-planner): Support to generate SQL top-k CountMinSketchWithHeap configs

### LINKED ISSUE #394: feat(asap-planner): Add support for SQL top-k config generation to asap_planner
## Summary

Extend `asap_planner` to generate streaming and inference configs for SQL top-k queries (`COUNT/SUM … GROUP BY … ORDER BY <alias> DESC LIMIT k`).

Today the planner treats these as plain `COUNT`/`SUM` and emits `DeltaSetAggregator` + `CountMinSketch` (two aggregations). The query engine expects a single self-keyed `CountMinSketchWithHeap` for top-k. 

This change closes that gap so top-k precompute configs can be produced from planner workload YAML, consistent with quantile, HLL, and MinMax workloads.

---

## Motivation

SQL top-k is already supported at query time: the engine detects the pattern via `detect_sql_topk`, promotes to `Statistic::Topk`, and resolves a `CountMinSketchWithHeap` with the right `count_events` weighting (COUNT vs SUM).

The planner had no equivalent path. For any SQL query with `ORDER BY … DESC LIMIT k`, it still planned a standard approximate COUNT/SUM pipeline:

- Two streaming aggs (`DeltaSetAggregator` + `CountMinSketch`)
- Subpopulation label layout (`aggregated: [group key]`, with a separate delta set)
- No `heapsize`, no `count_events`

That does not match what the engine needs for top-k execution.

Sharing detection logic in `sql_utilities` keeps planner output aligned with engine behavior and avoids two implementations drifting apart.

---

## Proposed behavior

When the planner parses a flat SQL query of this shape:

```sql
SELECT <key>, COUNT(<col>) AS <alias>   -- or SUM(<col>) AS <alias>
FROM <table>
WHERE <time window>
GROUP BY <key>
ORDER BY <alias> DESC
LIMIT k
```

It should:

1. Detect top-k via shared `detect_sql_topk()`:
   - Requires `LIMIT`
   - `ORDER BY` on aggregate alias (DESC)
   - Only supports `COUNT` or `SUM`

2. Plan `Statistic::Topk` instead of `Count`/`Sum`

3. Emit a single `CountMinSketchWithHeap` per query  
   - No `DeltaSetAggregator`

4. Use heap-only label layout:
   - grouping: []
   - aggregated: [<group key>]
   - rollup: [metadata columns not in GROUP BY]

5. Set sketch parameters:
   - `aggregationSubType`: `topk`
   - `heapsize`: `k × heap_multiplier` (default multiplier = 4)
   - `count_events`:  
     - `true` for COUNT  
     - `false` for SUM  

---

### Behavior Summary

| Property       | COUNT top-k            | SUM top-k              |
|----------------|------------------------|------------------------|
| Agg type       | CountMinSketchWithHeap | CountMinSketchWithHeap |
| Sub-type       | topk                   | topk                   |
| Streaming aggs | 1                      | 1                      |
| count_events   | true                   | false                  |
| Labels         | heap-only self-keyed   | heap-only self-keyed   |

---

### Notes

- Queries **without `ORDER BY + LIMIT`** continue using the existing CMS + DeltaSet path unchanged.
- PromQL `topk(k, …)` is unchanged.
- **Not supported**:
  - Nested SQL
  - Multiple aggregates in `SELECT`
  - Non-COUNT/SUM aggregates for top-k planning

akanksha-akkihal:
## Summary

Adds SQL top-k config generation to `asap_planner` and hoists shared detection into `sql_utilities`, so the planner and query engine use the same rules for `COUNT/SUM … GROUP BY … ORDER BY <alias> DESC LIMIT k`.

Previously the planner treated these queries as plain COUNT/SUM and emitted `DeltaSetAggregator` + `CountMinSketch`. The engine expects a single heap-only `CountMinSketchWithHeap`.
### Shared detection (`sql_utilities`)
- Added `TopkWeighting`, `SqlTopk`, and `detect_sql_topk()` to `sqlhelper.rs`
- Re-exported from `ast_matching/mod.rs`
- Single-layer detection only; nested-pattern gating stays with callers

### Planner (`asap-planner-rs`)
- Detects top-k via shared `detect_sql_topk()` and plans `Statistic::Topk`
- Emits one `CountMinSketchWithHeap` per query (no `DeltaSetAggregator`)
- Heap-only labels: `grouping: []`, `aggregated: [group key]`, `rollup: [other metadata]`
- Sketch params: `aggregationSubType: topk`, `heapsize = k × 4`, `count_events: true/false` for COUNT/SUM
- Extended `build_sketch_parameters` with optional `topk_count_events` (PromQL unchanged)
- Added `PlannerOutput::aggregation_parameter()` for tests
- Integration tests: COUNT top-k, SUM top-k, non-top-k fallback

### Query engine (`asap-query-engine`)
- Removed duplicate `detect_sql_topk` / types; imports shared implementation from `sql_utilities`
- Existing `detect_topk_tests` unchanged in behavior

## Behavior

| | Before | After |
|---|---|---|
| Flat COUNT/SUM + ORDER BY + LIMIT | CMS + DeltaSet (2 aggs) | Single `CountMinSketchWithHeap` |
| `count_events` | absent | `true` (COUNT) / `false` (SUM) |
| Nested SQL | planner error (`n != 1`) | same (still unsupported) |

PromQL `topk(k, …)` is unchanged. Plain COUNT/SUM without ORDER BY/LIMIT still uses the CMS + DeltaSet path.

### Planner integration tests (`asap-planner-rs/tests/sql_integration.rs`)

- **`spatial_count_topk_heap`** — COUNT + ORDER BY DESC + LIMIT → 1 agg, `CountMinSketchWithHeap` (`topk`); no CMS/DeltaSet; labels + `heapsize: 40`, `count_events: true`
- **`spatial_sum_topk_heap`** — SUM + ORDER BY DESC + LIMIT → same heap path; `count_events: false`
- **`spatial_count_without_order_by_is_not_topk`** — COUNT only → CMS + DeltaSet (2 aggs); no heap

### DIFF
--- a/asap-common/dependencies/rs/sql_utilities/src/ast_matching/mod.rs
+++ b/asap-common/dependencies/rs/sql_utilities/src/ast_matching/mod.rs
@@ -3,6 +3,6 @@ pub mod sqlparser_test;
 pub mod sqlpattern_matcher;
 pub mod sqlpattern_parser;
 
-pub use sqlhelper::{SQLSchema, Table};
+pub use sqlhelper::{detect_sql_topk, SQLSchema, SqlTopk, Table, TopkWeighting};
 pub use sqlpattern_matcher::*;
 pub use sqlpattern_parser::*;
--- a/asap-common/dependencies/rs/sql_utilities/src/ast_matching/sqlhelper.rs
+++ b/asap-common/dependencies/rs/sql_utilities/src/ast_matching/sqlhelper.rs
@@ -280,6 +280,10 @@ impl SqlTopk {
 /// query pattern type; the planner rejects nested queries up front).
 pub fn detect_sql_topk(query_data: &SQLQueryData) -> Option<SqlTopk> {
     let k = query_data.limit?;
+    // LIMIT 0 is an empty-result query, not a top-k heavy-hitter request.
+    if k == 0 {
+        return None;
+    }
     // Need a GROUP BY key to rank and an ORDER BY to define "top".
     if query_data.labels.is_empty() || query_data.order_by.is_empty() {
         return None;
@@ -300,8 +304,11 @@ pub fn detect_sql_topk(query_data: &SQLQueryData) -> Option<SqlTopk> {
     if primary.ascending {
         return None;
     }
-    if query_data.aggregation_alias.as_deref() != Some(primary.column.as_str()) {
+    // ORDER BY may differ only by identifier case for unquoted aliases.
+    let alias = query_data.aggregation_alias.as_deref()?;
+    if !alias.eq_ignore_ascii_case(primary.column.as_str()) {
         return None;
     }
     Some(SqlTopk { k, weighting })
 }
+
--- a/asap-common/dependencies/rs/sql_utilities/src/ast_matching/sqlhelper.rs
+++ b/asap-common/dependencies/rs/sql_utilities/src/ast_matching/sqlhelper.rs
@@ -311,4 +311,3 @@ pub fn detect_sql_topk(query_data: &SQLQueryData) -> Option<SqlTopk> {
     }
     Some(SqlTopk { k, weighting })
 }
-
--- a/asap-common/dependencies/rs/sql_utilities/src/ast_matching/sqlhelper.rs
+++ b/asap-common/dependencies/rs/sql_utilities/src/ast_matching/sqlhelper.rs
@@ -222,3 +222,86 @@ impl SQLQueryData {
             }
     }
 }
+
+/// How a top-k query weights each observation fed into the heavy-hitter sketch.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum TopkWeighting {
+    /// `COUNT(col)`: every matching row contributes weight 1, so the heap ranks
+    /// keys by event frequency (`count_events: true`).
+    Count,
+    /// `SUM(col)`: every matching row contributes weight = `col`, so the heap
+    /// ranks keys by summed value (`count_events: false`).
+    ///
+    /// Assumes **non-negative** summands: `CountMinSketch` is a frequency sketch
+    /// and cannot represent negative weights, so a `SUM` over a column that can
+    /// go negative would produce meaningless estimates.
+    Sum,
+}
+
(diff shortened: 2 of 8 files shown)


milindsrivastava1997 (commented):
1. detect_sql_topk(&qdata) vs &sql_query.query_data[0] — sql.rs:108

  let sql_topk = detect_sql_topk(&qdata);  // ← from SQLPatternParser
  let agg_info = &sql_query.query_data[0].aggregation_info;  // ← from SQLPatternMatcher
  let labels = &sql_query.query_data[0].labels;

  All other field reads come from sql_query.query_data[0] (pattern-matcher output), but detect_sql_topk runs on qdata (raw parser output). For n=1 spatial queries these should be structurally equivalent, but if SQLPatternMatcher ever strips limit/order_by (they're not
  part of matches_sql_pattern), this silently diverges.

  The choice of qdata is correct today (it retains limit/order_by reliably), but deserves a one-line comment explaining why — especially since the query engine passes query_data from the pattern matcher at its call site (line 575).

  ---
  2. spatial_sum_topk_heap test coverage gap — tests/sql_integration.rs

  The COUNT test (spatial_count_topk_heap) verifies everything: heapsize, count_events, grouping/aggregated/rollup labels, window size, no DeltaSet/CMS. The SUM test only checks count_events = false and the aggregation type. At minimum it should verify:

  assert_eq!(out.aggregation_parameter("CountMinSketchWithHeap", "heapsize").and_then(|v| v.as_u64()), Some(40));
  assert_eq!(out.aggregation_labels("CountMinSketchWithHeap", "grouping"), Vec::<String>::new());
  assert_eq!(out.aggregation_labels("CountMinSketchWithHeap", "aggregated"), vec!["srcip".to_string()]);

  ---
  3. Missing negative test: ORDER BY ... ASC LIMIT k

  detect_sql_topk correctly returns None when primary.ascending, but there's no integration test confirming that ORDER BY alias ASC LIMIT k falls through to the CMS + DeltaSet path. Worth adding alongside the existing spatial_count_without_order_by_is_not_topk case.

  ---

milindsrivastava1997 (approved):

PR #436, opened 2026-06-23
fix(precompute_engine): Use value_column from labels at ingest

### LINKED ISSUE #432: fix(precompute):  value_column ignored at ingest -` COUNT(DISTINCT col)` estimates cardinality of the wrong column
## Summary

For a query such as:

```sql
SELECT srcip, COUNT(DISTINCT dstip)
FROM netflow_table
GROUP BY srcip;
```

the planner, inference configuration, and streaming configuration are all generated correctly:

- Aggregation type: `HLL`
- Grouping keys: `[srcip]`
- `value_column`: `dstip`

However, the ingest → precompute path never reads `AggregationConfig.value_column`. Instead, it always feeds the scalar value that arrived on the wire (the dataset's configured value column, e.g. `pkt_len`) into the accumulator.

As a result, ASAP computes:

```sql
COUNT(DISTINCT pkt_len) GROUP BY srcip
```

while ClickHouse correctly computes:

```sql
COUNT(DISTINCT dstip) GROUP BY srcip
```

The two systems are answering different queries, making fidelity and benchmark comparisons invalid.

---

## Root Cause

Although `value_column` is propagated correctly through planning and configuration, it is never used during ingest or precomputation.

### JSON Ingest

`json_ingest.rs` reads only `JsonFileIngestConfig.value_col` (the dataset default value column).

### Remote Write Ingest

Remote-write ingestion forwards the Prometheus sample value directly:

```rust
Sample.value
```

### Routing

`ingest_source.rs` forwards:

```rust
(labels, ts, s.value)
```

for every aggregation and performs no per-aggregation `value_column` resolution.

### Worker Execution

`worker.rs::apply_sample()` hashes the supplied value into the accumulator and never consults `config.value_column`:

```rust
fn apply_sample(updater, series_key, val, ts, config) {
    if updater.is_keyed() {
        ...
        updater.update_keyed(&key, val, ts);
    } else {
        updater.update_single(val, ts);
    }
}
```

For HLL aggregations, `val` is therefore the dataset's wire value (e.g. `pkt_len`) rather than the requested column (`dstip`).

Consequently, the HLL accumulator estimates:

```sql
COUNT(DISTINCT pkt_len)
```

instead of:

```sql
COUNT(DISTINCT dstip)
```

---

### Correctness

`COUNT(DISTINCT col)` returns the cardinality of the dataset's configured value column rather than the column requested by the query.

### Benchmark Validity

Comparisons between ASAP and ClickHouse for cardinality queries are invalid because the systems are computing distinct counts over different columns.

### Silent Failure

No error or warning is emitted:

- The planner generates the expected configuration.
- `value_column` appears correctly populated.
- The query executes successfully.
- Results look plausible but are incorrect.

This makes the issue particularly difficult to detect and can lead to misleading fidelity and performance conclusions.

## Acceptance Criteria

- `COUNT(DISTINCT dstip) GROUP BY srcip` over the netflow workload yields ASAP cardinality matching ClickHouse `uniq(dstip)` within HLL tolerance.
- `SUM` / `quantile` / `top-k` queries are unaffected (still aggregate the wire value column).
- Regression test: two samples with the same `srcip`, different `dstip`, identical `pkt_len` → HLL estimate ≈ 2 (currently 1).

akanksha-akkihal:
## Summary

Fixes #432. The streaming config's `value_column` was ignored at ingest time, so every aggregation consumed the default wire scalar (e.g. `pkt_len`) regardless of what column it was configured to aggregate. The visible symptom: `COUNT(DISTINCT dstip)` made the HLL sketch estimate the cardinality of `pkt_len` instead of `dstip`, diverging sharply from the ClickHouse baseline (collapsing to ≈1 distinct value per group).

## Root cause

The wire format collapses each sample to a single scalar plus a set of labels carried in the series key. At ingest, `apply_sample()` always fed that wire scalar straight into the accumulator and never consulted `config.value_column`. For HLL `COUNT(DISTINCT dstip)`, `dstip` arrives as a **label**, not as the wire value, so it never reached the sketch.

## Changes

- **`resolve_sample_value(series_key, wire_val, config)`** (new, in `precompute_engine/worker.rs`): if `config.value_column` names a column present among the series labels, parse that label's value to `f64` and aggregate it; otherwise fall back to the wire value.
- **`apply_sample()`**: resolves the value via the helper before updating both keyed and non-keyed accumulators.

### Why this is general and safe
- Numeric value columns like `pkt_len` are never transmitted as labels, so SUM / quantile / top-k transparently keep using the wire scalar — no behavior change for existing aggregations.
- Only label-present distinct targets (e.g. `dstip`) get substituted, which is exactly the broken case.
- The store is keyed by `aggregation_id`, so nothing about lookup/routing changes.

### Known limitation (follow-up)
Non-numeric distinct targets (e.g. `proto`) can't be parsed to `f64`; they log a debug message and fall back to the wire value. A byte/string hashing path for non-numeric `COUNT(DISTINCT)` is left as a follow-up.

## Tests

- `resolve_sample_value_uses_label_when_value_column_present`
- `resolve_sample_value_falls_back_when_column_not_a_label`
- `resolve_sample_value_none_value_column_uses_wire_value`
- `resolve_sample_value_non_numeric_label_falls_back`
- `hll_counts_distinct_value_column_not_wire_value` — worker integration test: two samples, same `srcip`, different `dstip`, identical `pkt_len`; asserts the closed window's HLL estimate is ≈2 (would be 1 with the bug).

## Test plan

- [ ] `cargo test -p query_engine_rust resolve_sample_value`
- [ ] `cargo test -p query_engine_rust hll_counts_distinct`
- [ ] E2E: replay netflow, then compare `COUNT(DISTINCT dstip) GROUP BY srcip` between ASAP (`:8088/clickhouse/query`) and ClickHouse (`:8123`) over the same window — ASAP cardinalities should track ClickHouse within HLL error (~1–2% at precision 14) instead of collapsing to ≈1.

### DIFF
--- a/asap-query-engine/src/precompute_engine/worker.rs
+++ b/asap-query-engine/src/precompute_engine/worker.rs
@@ -9,7 +9,7 @@ use crate::precompute_engine::window_manager::WindowManager;
 use crate::precompute_operators::sum_accumulator::SumAccumulator;
 use asap_types::aggregation_config::AggregationConfig;
 use std::collections::{BTreeMap, HashMap};
-use std::sync::atomic::{AtomicBool, AtomicI64, AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicI64, AtomicUsize, Ordering};
 use std::sync::Arc;
 use tokio::sync::mpsc;
 use tracing::{debug, debug_span, info, warn};
@@ -914,9 +914,13 @@ fn apply_sample(
 /// The label value is parsed as `f64`. This is lossless for the netflow IPv4
 /// `u32` columns (≤ 2^32 < 2^53) and consistent with the replay's int→f64
 /// convention; cardinality only requires distinct inputs to map to distinct
-/// hashes, so the exact hash need not match the baseline engine. Non-numeric
-/// distinct targets (e.g. `proto`) are not yet supported — they fall back to
-/// `wire_val` (see the follow-up for a byte/string hashing path).
+/// hashes, so the exact hash need not match the baseline engine.
+///
+/// # Panics
+/// Panics if `value_column` names a label whose value is not numeric. Non-numeric
+/// distinct targets (e.g. `proto`) are not yet supported; this is a temporary
+/// guard until the path is refactored to return a `Result` (see follow-up for a
+/// byte/string hashing path).
 fn resolve_sample_value(
     labels: &HashMap<&str, &str>,
     wire_val: f64,
@@ -932,24 +936,15 @@ fn resolve_sample_value(
 
     match raw.parse::<f64>() {
         Ok(v) => v,
-        Err(_) => {
-            // Non-numeric distinct targets (e.g. COUNT(DISTINCT proto)) are not yet
-            // supported: falling back to the wire value silently produces an INCORRECT
-            // aggregate, so surface it as a warning. Warn-once (process-global guard)
-            // keeps this off the per-sample hot path — otherwise a single misconfigured
-            // aggregation would emit a warning for every ingested sample.
-            static NON_NUMERIC_WARNED: AtomicBool = AtomicBool::new(false);
-            if !NON_NUMERIC_WARNED.swap(true, Ordering::Relaxed) {
-                warn!(
-                    "value_column '{}' label value {:?} is not numeric; falling back to the \
-                     wire value. Non-numeric distinct targets (e.g. COUNT(DISTINCT proto)) are \
-                     not yet supported and will produce INCORRECT results. \
-                     (This warning is logged once per process.)",
-                    col, raw
-                );
-            }
-            wire_val
-        }
+        // Non-numeric distinct targets (e.g. COUNT(DISTINCT proto)) are not yet
+        // supported: silently falling back to the wire value would produce an
+        // INCORRECT aggregate, so fail loudly instead. This panic is a temporary
+        // measure — the longer-term fix is to make this path return a `Result`
+        // and propagate the error up to the caller.
+        Err(_) => panic!(
+            "value_column '{col}' label value {raw:?} is not numeric; non-numeric distinct \
+             targets (e.g. COUNT(DISTINCT proto)) are not yet supported"
+        ),
     }
 }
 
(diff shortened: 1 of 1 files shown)


akanksha-akkihal:
parse series labels once per sample; warn on non-numeric value_column

Follow-up to the value_column ingest fix. Two improvements in `apply_sample`:

- Parse the series labels at most once and share the map between value
  resolution and aggregated-key extraction. Previously the keyed path
  parsed labels twice per sample, and the non-keyed path with a
  `value_column` parsed once even when not needed. Now labels are parsed
  only when the accumulator is keyed or a `value_column` is set, removing
  redundant allocations from the per-sample ingest hot path.
- Surface non-numeric `value_column` targets (e.g. `COUNT(DISTINCT proto)`)
  as a warn-once instead of a per-sample debug log, since falling back to
  the wire value silently produces incorrect results. A process-global
  `AtomicBool` guard keeps the warning off the hot path.

`resolve_sample_value` and `extract_aggregated_key_from_series` now take the
pre-parsed label map; tests updated accordingly.

asap-query-engine/src/precompute_engine/worker.rs
```diff
@@ -872,22 +872,94 @@ fn apply_sample(
...
+            // keeps this off the per-sample hot path — otherwise a single misconfigured
+            // aggregation would emit a warning for every ingested sample.
+            static NON_NUMERIC_WARNED: AtomicBool = AtomicBool::new(false);
+            if !NON_NUMERIC_WARNED.swap(true, Ordering::Relaxed) {
+                warn!(
+                    "value_column '{}' label value {:?} is not numeric; falling back to the \
+                     wire value. Non-numeric distinct targets (e.g. COUNT(DISTINCT proto)) are \
+                     not yet supported and will produce INCORRECT results. \
+                     (This warning is logged once per process.)",
+                    col, raw
```
milindsrivastava1997: Pls change this to a panic! for now. 
Ultimately, we should propagate an Error and have the return type as `Result`

milindsrivastava1997 (changes requested):

milindsrivastava1997 (approved):

PR #450, opened 2026-06-26
fix(planner): change more variables to ms precision

### LINKED ISSUE #427: feat(planner): millisecond precision for SQL/Elastic data_ingestion_interval
## Background

#398 adds millisecond precision to the PromQL planner path (`prometheus_scrape_interval`, `step`, `range_duration`, `repetition_delay`, `window_size`, `slide_interval`). The SQL and Elastic controllers have an analogous seconds-based field, `data_ingestion_interval` (`asap-planner-rs/src/sql/generator.rs`, `src/elastic_dsl/generator.rs`), with the same underlying problem — sub-second ingestion intervals can't be expressed.

The original issue (#398) doesn't mention this field, so it was explicitly left out of scope there and flagged as a known follow-up rather than silently expanded into.

## Work

Same fix shape as #398: rename `data_ingestion_interval` to `data_ingestion_interval_ms` across the SQL/Elastic generator path, delete now-redundant `* 1000` conversions at consumption sites, update fixtures/tests.

See `.design_docs/issue-398-millisecond-precision-plan.md` ("Scope boundary" and "Follow-up Issues" sections) for context.

milindsrivastava1997:


milindsrivastava1997:
Merged into #449 . Closing.

PR #480, opened 2026-06-29
feat(asap-tools): add remote_monitor (CPU/memory) support to experiment_run_clickhouse.py

### LINKED ISSUE #479: feat(asap-tools): add remote_monitor (CPU/memory) support to `experiment_run_clickhouse.py`
## Summary

`experiment_run_clickhouse.py` had no resource-monitoring support, so ClickHouse/SQL experiments never produced `monitor_output.json` and `analyze_monitor_output.py` had nothing to read. Only `experiment_run_e2e.py` wired in `RemoteMonitorService`. This adds the same capability to the ClickHouse runner behind a config flag, generalizing `remote_monitor.py` / `RemoteMonitorService` to support a ClickHouse (non-Prometheus) backend.

## Motivation

Latency and query-accuracy comparisons (`compare_latencies.py`, `compare_query_results.py`) work for ClickHouse experiments, but CPU/memory profiling did not, because the runner never launched `remote_monitor.py`. Resource cost is needed to compare baseline (ClickHouse) vs sketchdb at different data rates.

## Acceptance criteria

- [ ] `flow.remote_monitor=true` produces `monitor_output.json` for both baseline and sketchdb.
- [ ] `analyze_monitor_output.py` consumes it and emits CPU/memory stats + plots.
- [ ] `flow.remote_monitor=false` (default) leaves existing behavior unchanged.
- [ ] No regression in `experiment_run_e2e.py` (shared `RemoteMonitorService` / `remote_monitor.py`).

## Related

- Mirrors monitoring wiring in `experiment_run_e2e.py`.
- Complements existing ClickHouse latency/query-result comparison tooling.

akanksha-akkihal:
## Changes

**`experiment_run_clickhouse.py`**
- New `flow.remote_monitor` flag gates monitoring.
- `_run_query_workload()` wraps the prometheus-client invocation in `RemoteMonitorService` when enabled; otherwise unchanged `_run_query_client()`.
- `_clickhouse_monitor_keywords()` selects ps-matchable process keywords: `clickhouse` (baseline) and `query_engine_rust` (sketchdb).
- Waits for the timed monitor to finish before teardown.

**`config/config.yaml`**
- Added `flow.remote_monitor: false` (default off).

**`RemoteMonitorService.start()`**
- New `backend_type` and `monitor_keywords` parameters.
- When `monitor_keywords` is provided, use it directly (skip Prometheus-config keyword discovery).
- Invoke `python3.11` (deps live there on the node) instead of `python3`.
- Pass `--backend_type` through to `remote_monitor.py`.
- `arroyo_service` made optional.

**`remote_monitor.py`**
- New `--backend_type` arg, forwarded to `PrometheusClientService` (was hardcoded `"prometheus"`); `clickhouse` skips PromQL query-time alignment.
- Missing keyword → warn and skip instead of raising, so monitoring tolerates absent processes.

## Issues fixed along the way

- `python3` on the node lacked `loguru`/experiment deps → switched to `python3.11`.
- `clickhouse-server` container name didn't match `ps` (host-network bare-metal ClickHouse) → use `clickhouse` keyword.
- A defaulted parameter preceded required params in `start()` → reordered.

## Usage

```bash
python3 experiment_run_clickhouse.py \
  experiment_type=clickhouse experiment.name=netflow_2mrps \
  providers.cloudlab.num_nodes=1 providers.cloudlab.username=<user> \
  providers.cloudlab.hostname_suffix=<suffix> \
  use_container.prometheus_client=false use_container.query_engine=false \
  use_container.controller=false \
  flow.remote_monitor=true
```

Then:

```bash
python3 post_experiment/analyze_monitor_output.py \
  experiment_outputs/<name>/<mode>/remote_monitor_output/monitor_output.json \
  --print --plot --save --save-to-experiment-dir
```

## Output

```
experiment_outputs/<name>/<baseline|sketchdb>/remote_monitor_output/monitor_output.json
```

### DIFF
--- a/asap-tools/experiments/config/config.yaml
+++ b/asap-tools/experiments/config/config.yaml
@@ -47,7 +47,6 @@ flow:
   no_teardown: false
   replace_query_engine_with_dumb_consumer: false
   steady_state_wait: 60
-  remote_monitor: false  # ClickHouse runner: collect CPU/memory via remote_monitor.py
   skip_copy_prometheus_data: false  # Skip copying Prometheus data back to local machine
 
 
--- a/asap-tools/experiments/config/config.yaml
+++ b/asap-tools/experiments/config/config.yaml
@@ -47,6 +47,7 @@ flow:
   no_teardown: false
   replace_query_engine_with_dumb_consumer: false
   steady_state_wait: 60
+  remote_monitor: false  # ClickHouse runner: collect CPU/memory via remote_monitor.py
   skip_copy_prometheus_data: false  # Skip copying Prometheus data back to local machine
 
 
--- a/asap-tools/experiments/experiment_run_clickhouse.py
+++ b/asap-tools/experiments/experiment_run_clickhouse.py
@@ -185,10 +185,6 @@ def _run_query_workload(
         controller_remote_output_dir=controller_remote_output_dir,
         use_container_prometheus_client=use_container,
         prometheus_client_parallel=parallel,
-        # monitoring_tool only selects the Prometheus/VictoriaMetrics scrape-config
-        # keyword, which is unused on the ClickHouse path: process keywords are
-        # chosen from backend_type="clickhouse" below.
-        monitoring_tool="clickhouse",
         backend_type="clickhouse",
     )
     if not manual_remote_monitor and constants.AVOID_REMOTE_MONITOR_LONG_SSH:
--- a/asap-tools/experiments/experiment_run_clickhouse.py
+++ b/asap-tools/experiments/experiment_run_clickhouse.py
@@ -153,7 +153,6 @@ OmegaConf.register_new_resolver(
 
 
 def _run_query_workload(
-    *,
     experiment_mode: str,
     experiment_output_dir: str,
     controller_client_config: str,
@@ -186,6 +185,9 @@ def _run_query_workload(
         controller_remote_output_dir=controller_remote_output_dir,
         use_container_prometheus_client=use_container,
         prometheus_client_parallel=parallel,
+        # monitoring_tool only selects the Prometheus/VictoriaMetrics scrape-config
+        # keyword, which is unused on the ClickHouse path: process keywords are
+        # chosen from backend_type="clickhouse" below. Harmless placeholder.
         monitoring_tool="prometheus",
         backend_type="clickhouse",
     )
--- a/asap-tools/experiments/experiment_run_clickhouse.py
+++ b/asap-tools/experiments/experiment_run_clickhouse.py
@@ -185,7 +185,7 @@ def _run_query_workload(
         controller_remote_output_dir=controller_remote_output_dir,
         use_container_prometheus_client=use_container,
         prometheus_client_parallel=parallel,
-        backend_type="clickhouse",
+        backend_protocol="clickhouse",
     )
     if not manual_remote_monitor and constants.AVOID_REMOTE_MONITOR_LONG_SSH:
         remote_monitor_service.wait_for_remote_monitor_to_finish(
--- a/asap-tools/experiments/experiment_run_clickhouse.py
+++ b/asap-tools/experiments/experiment_run_clickhouse.py
@@ -109,8 +109,10 @@ from experiment_utils.services import (
 )
 from experiment_utils.services.misc import ControllerService, DiscoveryBackend
 from experiment_utils.services.query_engine import QueryEngineRustService
+from experiment_utils.services.remote_monitor_service import RemoteMonitorService
 
 CLICKHOUSE_DATABASE = "default"
+REMOTE_PROCESS_POLLING_INTERVAL = 10
 
 
 def _inline_sql_queries_in_experiment_config(local_experiment_root_dir: str) -> None:
@@ -207,7 +209,7 @@ def _run_query_client(
         )
     else:
         cmd = (
-            f"python3 -u main_prometheus_client.py"
+            f"python3.11 -u main_prometheus_client.py"
             f" --config_file {config_file}"
             f" --output_dir {output_dir}"
             f" --output_file prometheus_client_output.txt"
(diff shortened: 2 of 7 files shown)


asap-tools/experiments/remote_monitor.py
```diff
@@ -516,6 +520,12 @@ def main(args):
         required=True,
         help="Streaming engine type (e.g. precompute, arroyo)",
     )
+    parser.add_argument(
+        "--backend_type",
+        type=str,
+        default="prometheus",
```
milindsrivastava1997: Pls remove the default here

asap-tools/experiments/config/config.yaml
```diff
@@ -47,6 +47,7 @@ flow:
   no_teardown: false
   replace_query_engine_with_dumb_consumer: false
   steady_state_wait: 60
+  remote_monitor: false  # ClickHouse runner: collect CPU/memory via remote_monitor.py
```
milindsrivastava1997: We can remove this flag, and just have remote_monitor on by default.

asap-tools/experiments/experiment_utils/services/remote_monitor_service.py
```diff
@@ -43,12 +43,14 @@ def start(
...
         streaming_engine: str,
         query_engine_service: "BaseQueryEngineService",
-        arroyo_service: "ArroyoService",
+        arroyo_service: Optional["ArroyoService"],
         controller_remote_output_dir: str,
         use_container_prometheus_client: bool,
         prometheus_client_parallel: bool,
         monitoring_tool: str,
         timed_duration: Optional[int] = None,
+        backend_type: str = "prometheus",
```
milindsrivastava1997: Remove the default here.

asap-tools/experiments/experiment_utils/services/remote_monitor_service.py
```diff
@@ -169,7 +185,7 @@ def start(
                     keywords.append("arroyo.*worker")
 
         cmd = (
-            "python3 -u remote_monitor.py "
+            "python3.11 -u remote_monitor.py "
```
milindsrivastava1997: loguru should have support for python3 that's already there on Cloudlab. Didn't understand why we need to switch these invocations to python3.11?

asap-tools/experiments/experiment_run_clickhouse.py
```diff
@@ -224,6 +226,81 @@ def _run_query_client(
...
 
+def _clickhouse_monitor_keywords(experiment_mode: str) -> list[str]:
+    """Process keywords for remote_monitor on ClickHouse experiments.
+
+    Use ps-friendly patterns (not Docker container names): host-network
+    ClickHouse does not show up as ``clickhouse-server`` in ``ps``, and
+    ``docker inspect`` can fail in background SSH shells.
+    """
+    if experiment_mode == constants.SKETCHDB_EXPERIMENT_NAME:
+        return [constants.QUERY_ENGINE_RS_PROCESS_KEYWORD]
```
milindsrivastava1997: This logic is inconsistent with that at remote_monitor_service.py L 90-93
For the time being, can you put clickhouse keywords also in remote_monitor_service.py? and remove the `monitor_keywords` argument? I want them to be in one place, even if they're messy for the time being.
I will figure out a cleaner abstraction for this.

asap-tools/experiments/experiment_run_clickhouse.py
```diff
@@ -224,6 +226,81 @@ def _run_query_client(
...
+        arroyo_pids=None,
+        manual_mode=manual_remote_monitor,
+        do_local_flink=False,
+        streaming_engine="precompute",
+        query_engine_service=query_engine_service,
+        arroyo_service=None,
+        controller_remote_output_dir=controller_remote_output_dir,
+        use_container_prometheus_client=use_container,
+        prometheus_client_parallel=parallel,
+        monitoring_tool="prometheus",
```
milindsrivastava1997: should this be "prometheus" ?

milindsrivastava1997 (changes requested):

asap-tools/experiments/experiment_run_clickhouse.py
milindsrivastava1997: @akanksha-akkihal This is unaddressed

asap-tools/experiments/experiment_run_clickhouse.py
```diff
@@ -150,78 +152,48 @@ def _expand_groups(groups):
 )
 
 
-def _run_query_client(
-    provider,
-    node_offset: int,
-    config_file: str,
-    output_dir: str,
+def _run_query_workload(
+    *,
```
milindsrivastava1997: Why do we need a `*` here?

asap-tools/experiments/experiment_run_clickhouse.py
akanksha-akkihal: `monitoring_tool` is only used in `RemoteMonitorService.start()` to pick the
scrape-config keyword. On the ClickHouse path that branch is never used — process
keywords come from `backend_type == "clickhouse"` (→ `clickhouse` for baseline,
`query_engine_rust` for sketchdb), and `config_keywords` is ignored. I've changed
the value from `"prometheus"` to `"clickhouse"` so it's not misleading, and added
an inline comment. (`"clickhouse"` falls into the existing `else` branch, so
behavior is unchanged.)

milindsrivastava1997: Is there a difference between monitoring_tool and backend_type?

akanksha-akkihal: Yes, `monitoring_tool` and `backend_type` are different:

- **`monitoring_tool`** (`prometheus` / `victoriametrics`) — which **metrics stack**
  scrapes/stores CPU/mem/latency. Its only job is to pick the scrape-config process
  keyword so the monitor can find that stack's process.
- **`backend_type`** (`prometheus` / `clickhouse`) — which **data store the queries
  run against**. Forwarded to the prometheus-client so it forms PromQL vs SQL.

In normal ASAP runs both are `prometheus` (it's both the metrics stack and the query
backend), so they were never clearly separated. ClickHouse is the first case where
they split.

So, on whether we need `monitoring_tool`:

- **Prometheus experiments**: needed — it selects the scrape-config keyword
  (`prometheus.yml` / vmagent) so the monitor knows which metrics-stack process to sample.
- **ClickHouse experiments**: not needed — there's no Prometheus/VM scrape stack, and the
  monitored processes (`clickhouse` / `query_engine_rust`) are chosen by
  `backend_type="clickhouse"` instead. So `monitoring_tool` is just an unused placeholder here.

milindsrivastava1997: This explanation doesn't make sense, not sure I understand the difference. Aren't both of them used to determine what keywords to monitor?

akanksha-akkihal: Yeah, I see that both of them determine what keywords to monitor. I could potentially remove `monitoring_tool` and just keep `backend_type` but problem child is `victoriametrics`. The e2e runner passes `monitoring_tool=cfg.experiment_params.monitoring.tool`, which can legitimately be `victoriametrics`  and that's the only thing that selects the vmagent/vmsingle scrape keywords (lines 70–72). backend_type has no victoriametrics value, so if I dropped monitoring_tool and keyed everything off backend_type, a VictoriaMetrics run would fall through to the Prometheus keyword and monitor the wrong collector.

akanksha-akkihal: So, I just kept both because I wasn't sure about the Victoria metrics path.

milindsrivastava1997: "backend_type has no victoriametrics value," - why is this not possible?

milindsrivastava1997:
Follow-up cleanup on the variable names introduced here:

- `monitoring_tool` renamed to `backend_tool`: it describes which *concrete tool* is running on the node for monitoring ("prometheus" or "victoriametrics"), which controls which process keywords `remote_monitor_service` searches for.
- `backend_type` renamed to `backend_protocol`: it describes which *query protocol* the system under test speaks ("prometheus" = PromQL, "clickhouse" = SQL), which controls whether the prometheus-client applies `--align-query-time`.

The two params are related (both about the backend stack) but not the same axis — the shared `backend_` prefix signals the relationship, while `_tool` vs `_protocol` signals the distinction. On the ClickHouse path `backend_tool` is now `Optional` (defaulting to `None`) since no TSDB runs there and the value was previously a dead dummy `"clickhouse"`.

milindsrivastava1997 (approved):

PR #483, opened 2026-06-30
fix(precompute): correctly parse configuration parameters from planner for CMS with heap

### LINKED ISSUE #482: fix(precompute): cms_params() rejects planner-emitted depth/width params for CountMinSketch
Sub-issue of #477 (defect 1 only).

### Problem

`cms_params()` in `accumulator_factory.rs` only reads `row_num`/`col_num`, but the planner (`sketch.rs`) emits `depth`/`width` for `CountMinSketch`. Every plain-CMS aggregation throws:

```
WARN ... CMS config missing required parameter: row_num
```

and never ingests. Regression from #397, which made params required without adding the `depth`/`width` alias on this path.

`cms_heap_params()` (for `CountMinSketchWithHeap`) already accepts `depth`/`width` first — the fix is to mirror that pattern in `cms_params()`.

### Fix

In `cms_params()`, try `depth`/`width` before `row_num`/`col_num` (matching `cms_heap_params()`).

milindsrivastava1997:


asap-query-engine/src/precompute_engine/accumulator_factory.rs
```diff
@@ -1336,10 +1341,46 @@ mod tests {
...
+    fn test_cms_depth_width_params_accepted() {
+        // Planner emits depth/width; engine must accept them without error.
+        use std::collections::HashMap;
+        let mut params = HashMap::new();
+        params.insert("depth".to_string(), serde_json::json!(3_u64));
+        params.insert("width".to_string(), serde_json::json!(1024_u64));
+        let config = AggregationConfig::new(
+            21,
+            AggregationType::CountMinSketch,
+            "sum".to_string(),
```
zzylol: What is this field for?

asap-query-engine/src/precompute_engine/accumulator_factory.rs
```diff
@@ -711,19 +711,24 @@ fn kll_k_param(config: &AggregationConfig) -> Result<u16, String> {
...
-        as usize;
+    let read = |names: &[&str]| -> Result<usize, String> {
+        names
+            .iter()
+            .find_map(|n| config.parameters.get(*n).and_then(|v| v.as_u64()))
+            .map(|v| v as usize)
+            .ok_or_else(|| {
+                format!(
+                    "CMS config missing required parameter (tried: {})",
+                    names.join(", ")
```
zzylol, line 726: Could you give a yaml file example of the expected CMS config?

zzylol (changes requested):

asap-query-engine/src/precompute_engine/accumulator_factory.rs
milindsrivastava1997: This was used to distinguish whether CMS is doing sum vs count (i.e. `+= value` or `+= 1`)
However, I now see that it's not being used. Probably a bug crept in somewhere. Will remove this for now, and fix bug later.
Related https://github.com/ProjectASAP/ASAPQuery/issues/503

milindsrivastava1997: Filed in #505

asap-query-engine/src/precompute_engine/accumulator_factory.rs
milindsrivastava1997: The planner emits configs like:

```yaml
aggregationId: 1
aggregationType: CountMinSketch
aggregationSubType: ''
parameters:
  depth: 3
  width: 1024
```

asap-query-engine/src/precompute_engine/accumulator_factory.rs
zzylol: I think we don't need this field. Let's say CMS insert interface (key, value), you can assign value=1 or value = value from samples.

milindsrivastava1997: Yes, that's the insert interface, but the `AggregationConfig` data structure needs some way to distinguish those, since both those methods are valid for CMS.

zzylol: I'm not sure it should be in which layer, but the fundamental config here is to tell the computation engine which field to aggregate into CMS.

milindsrivastava1997: Are you saying there should be a dummy field with value = 1, if we want CMS to be used for count-based queries?

zzylol: yes

milindsrivastava1997: Sounds good. I will add support for that when I fix #505, and keep the implementation as it is for now. Is that fine?

zzylol: yes

zzylol (approved):

PR #494, opened 2026-07-01
refactor: slowly phase out QueryPatternType and QueryType enums

### LINKED ISSUE #487: QueryPatternType / QueryType::Spatial enums duplicate branching across 3 crates
### Summary
The temporal-vs-spatial distinction is encoded as two separate enums (`QueryPatternType` in `promql_utilities`, `QueryType::Spatial` in the SQL planner) that don't correspond to anything in SQL or Elastic, where a time range is always just a time range. Code branches on these enums in 8+ places across `promql_utilities`, `asap-planner-rs`, and `asap-query-engine`, duplicating logic that could instead be derived structurally from the parsed query.

### Problems this causes
- `get_statistics_to_compute`, `get_spatial_aggregation_output_labels`, and others take `QueryPatternType` as a parameter purely to re-derive information already present in the parsed match result (e.g. whether a range duration or aggregation op is present).
- `controller_patterns: HashMap<QueryPatternType, Vec<PromQLPattern>>` in `asap-query-engine/src/engines/simple_engine/mod.rs:140` iterates a `HashMap` keyed by the enum — iteration order is non-deterministic, and which pattern matches first can depend on that order.
- Every new query shape needs the enum-branching logic updated in multiple places (`promql.rs`, `sql.rs`, `mod.rs` in the query engine, plus the planner side) instead of one place.

### Evidence
- `promql_utilities/src/query_logics/enums.rs` — `QueryPatternType` definition
- `asap-query-engine/src/engines/simple_engine/mod.rs:140` — HashMap-keyed pattern registry
- `asap-query-engine/src/engines/simple_engine/promql.rs`, `.../sql.rs` — enum match sites
- `asap-planner-rs/src/planner/window.rs`, `cleanup.rs` — enum match sites in the planner

### Dependency
Cleaning this up depends on #486 — time ranges need to be explicit before the enum can be safely removed; otherwise the implicit-default logic the enum currently encodes has nowhere to go.

milindsrivastava1997:
- **refactor: start_timestamp logic for OnlySpatial**
- **updated start/end timestamp validation logic**
- **removed special build_spatiotemporal_context function**

### DIFF
--- a/asap-common/dependencies/rs/sql_utilities/src/ast_matching/sqlparser_test.rs
+++ b/asap-common/dependencies/rs/sql_utilities/src/ast_matching/sqlparser_test.rs
@@ -99,7 +99,7 @@ mod tests {
         if let Some(query_data) = SQLPatternParser::new(&schema, time).parse_query(&statements) {
             let result = matcher.query_info_to_pattern(&query_data);
             assert!(result.is_valid());
-            assert_eq!(result.query_type, vec![QueryType::Spatial]);
+            assert_eq!(result.query_type, vec![QueryType::TemporalGeneric]);
         }
     }
 
@@ -127,7 +127,7 @@ mod tests {
     fn test_dated_spatial_avg() {
         check_query(
             "SELECT AVG(value) FROM cpu_usage WHERE time BETWEEN DATEADD(s, -1, '2025-10-01 00:00:00') AND '2025-10-01 00:00:00' GROUP BY L1, L2, L3, L4",
-            vec![QueryType::Spatial],
+            vec![QueryType::TemporalGeneric],
             None,
         );
     }
@@ -136,7 +136,7 @@ mod tests {
     fn test_dated_spatial_quantile() {
         check_query(
             "SELECT QUANTILE(0.95, value) FROM cpu_usage WHERE time BETWEEN DATEADD(s, -1, '2025-10-01 00:00:00') AND '2025-10-01 00:00:00' GROUP BY L1",
-            vec![QueryType::Spatial],
+            vec![QueryType::SpatioTemporal],
             None,
         );
     }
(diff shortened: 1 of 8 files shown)


milindsrivastava1997:
### `QueryType::Spatial` phase-out

As of the latest commits on this branch, `query_info_to_pattern` no longer constructs `QueryType::Spatial`. It's replaced by the same two buckets every other query duration already used:

- **`SpatioTemporal`** — GROUP BY doesn't cover all metadata columns (partial label coverage)
- **`TemporalGeneric` / `TemporalQuantile`** — GROUP BY covers all metadata columns (full coverage), split by whether the aggregation is `QUANTILE`

Previously, a query spanning exactly one scrape interval was forced into `Spatial` regardless of its GROUP BY, while longer-duration queries went through a label-coverage check to decide between `TemporalGeneric`/`TemporalQuantile`/`SpatioTemporal`. That asymmetry is gone now — `Spatial` was never a real structural distinction, just `Temporal` with `range == scrape_interval`, so every query is classified the same way regardless of duration. Concretely: a 1-interval query with a full GROUP BY now lands in `TemporalGeneric`/`TemporalQuantile` (previously `Spatial`), and a 1-interval query with a partial GROUP BY now lands in `SpatioTemporal` (previously also `Spatial`, which silently hid the partial-label case).

The `Spatial` variant itself stays defined (dead code) until a later commit on this branch deletes it, once `sql.rs`'s dispatch is migrated off matching on it directly.

milindsrivastava1997:
### \`QueryPatternType\` phase-out, step 1: \`get_statistics_to_compute\`

Continuing the phase-out (see the earlier comment on \`QueryType::Spatial\`) — this time on the PromQL side's \`QueryPatternType\` enum (\`promql_utilities\`).

\`get_statistics_to_compute\` used to take \`pattern_type: QueryPatternType\` purely to decide which token to read:

\`\`\`rust
match pattern_type {
    OnlyTemporal | OneTemporalOneSpatial => read the function token (rate/increase/*_over_time),
    OnlySpatial => read the aggregation token (sum/avg/...),
}
\`\`\`

Replaced with direct token inspection, no enum needed:

\`\`\`rust
match_result.get_function_name()...or_else(|| match_result.get_aggregation_op()...)
\`\`\`

i.e. try the function token first (temporal wins when both a function and a spatial aggregation are present — same precedence as before for the "one temporal one spatial" shape), fall back to the aggregation token (the bare-aggregation "spatial" shape, e.g. \`sum(metric)\`). Behavior is identical; the enum branch and the token-presence check were doing the same thing, since a query can only reach one branch or the other depending on which tokens actually matched in the first place.

\`StatisticExtractionError::MissingStatistic\` correspondingly drops its \`pattern_type\` field (was only used for the error message) — it's now a unit variant.

PR #514, opened 2026-07-07
fix(tools): pin TZ=UTC across for ClickHouse experiments

milindsrivastava1997:
## Summary
- Pin `TZ=UTC` in the ClickHouse, asap-planner, and query-engine docker-compose configs and remote-launch commands so naive datetime-string time literals in SQL queries parse identically everywhere (previously could silently disagree by the shell's local UTC offset vs. ClickHouse's UTC default).
- Pass optional `sketch_parameters` through from experiment config into the generated SQL planner input YAML.
- Rsync `init_sql_file` to the remote node before executing it, instead of assuming it's already present there.

### DIFF
--- a/asap-planner-rs/docker-compose.yml.j2
+++ b/asap-planner-rs/docker-compose.yml.j2
@@ -2,6 +2,11 @@ services:
   controller:
     image: sketchdb-controller:latest
     container_name: {{ container_name }}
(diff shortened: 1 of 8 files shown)


milindsrivastava1997:
Split into #552 and #553

PR #522, opened 2026-07-12
feat(asap-tools): separate clickhouse ingest and query cpu/memory monitoring in `experiment run clickhouse`

### LINKED ISSUE #521: feat(asap-tools): Separate ClickHouse ingest and query CPU/memory monitoring in `experiment_run_clickhouse`
## Summary

Baseline ClickHouse experiments do not cleanly separate resource usage by phase: monitoring typically wraps only the query client, so **bulk ingest CPU/memory is missing **. For fair comparison against SketchDB, we need **distinct** monitoring for:

1. **Bulk ingest** (JSON load into ClickHouse)
2. **Query** (client workload against ClickHouse)

SketchDB already gets a continuous precompute+query monitor with per-thread attribution. Baseline should expose a comparable **ingest-only** signal via a dedicated monitor output.

akanksha-akkihal:
## Summary
- Separate **ClickHouse bulk-ingest** CPU/memory monitoring from the **query-phase** monitor in `experiment_run_clickhouse`.
- Baseline ingest writes `baseline/remote_monitor_output/monitor_output_ingest.json` and stops via a stop file (with kill fallback); query phase still writes `monitor_output.json`.
- SketchDB keeps a single continuous monitor that waits `flow.steady_state_wait` before the query client so precompute + query land in one time series (with bare-metal thread attribution).
- Harden stop/matching: poll during process walks, escalate stuck samplers with `kill()`, match `remote_monitor.py` by `--execution_mode` + experiment dir, and avoid writing `null` monitor JSON on timeout.


asap-tools/experiments/experiment_utils/services/remote_monitor_service.py
```diff
@@ -50,6 +52,8 @@ def start(
         backend_protocol: str,
         backend_tool: Optional[str] = None,
         timed_duration: Optional[int] = None,
+        pre_query_wait_seconds: int = 0,
+        monitor_interval_seconds: float = 1.0,
```
milindsrivastava1997, line 56: If this is defined in config.yaml, can we remove the default argument here

asap-tools/experiments/experiment_utils/services/remote_monitor_service.py
```diff
@@ -277,13 +488,19 @@ def wait_for_remote_monitor_to_finish(
         self,
         minimum_experiment_running_time: int,
         polling_interval: int = 10,
+        timeout: int = 600,
```
milindsrivastava1997, line 491: do we need this default arg?

asap-tools/experiments/classes/process_monitor.py
```diff
@@ -257,14 +272,19 @@ def start_monitor(
     return monitor, control_pipe, monitor_pipe
 
 
-def stop_monitor(monitor, control_pipe, monitor_pipe):
+def stop_monitor(monitor, control_pipe, monitor_pipe, timeout=30):
```
milindsrivastava1997, line 275: do we need this default arg

asap-tools/experiments/experiment_utils/services/remote_monitor_service.py
```diff
@@ -50,6 +52,8 @@ def start(
         backend_protocol: str,
         backend_tool: Optional[str] = None,
         timed_duration: Optional[int] = None,
+        pre_query_wait_seconds: int = 0,
```
milindsrivastava1997, line 55: do we need this default arg if the cmd line args already have a default value

asap-tools/experiments/remote_monitor.py
```diff
@@ -526,6 +595,26 @@ def main(args):
         required=True,
         help="Query backend for prometheus-client (prometheus or clickhouse)",
     )
+    parser.add_argument(
+        "--monitor_interval_seconds",
+        type=float,
+        default=1.0,
```
milindsrivastava1997, line 601: if this is defined in config.yaml, do we need this default arg?

milindsrivastava1997 (changes requested):
Does experiment_run_e2e.py need any changes? Has that script been run once with the new remote monitor to verify?

milindsrivastava1997:
Comments from CR with Claude:

## Hard violations
  1. remote_monitor.py — the new ingest execution-mode installs global SIGTERM/SIGINT handlers that just set a threading.Event() and never exit/restore. This means the SIGTERM the PR's own new kill-escalation path sends to stuck samplers gets swallowed, so the "force
  kill" safety net it adds can itself hang.
  2. remote_monitor_service.py::wait_for_remote_monitor_to_finish — gained an unconditional timeout=600 + force-kill escalation with no execution_mode/dir filter by default. This method is also called from the untouched experiment_run_e2e.py, so its "wait forever"
  semantics silently become "pkill any process matching remote_monitor.py after 10 min" for callers outside this PR's scope.

##  Magic numbers
  3. constants.py was touched (added INGEST_MONITOR_STOP_FILE) but the new timeout/poll literals (30, 60, 300, 10, 5, 2, 0.5s) scattered across the three other files stay inline, breaking the file's existing pattern of named interval constants.
  4. process_monitor.stop_monitor(..., timeout=30) only threads timeout into the first poll; the two .join() calls hardcode 10 and the kill escalation hardcodes 5 — inconsistent within one new function.

##  Implemented but wrong
5.  Manual mode is broken for the ingest monitor. _run_clickhouse_ingest_with_monitor unconditionally calls wait_for_remote_monitor_start(...) right after start_clickhouse_ingest_monitor(..., manual_mode=...), but unlike existing manual-mode call sites in this file
  (which guard with if not manual_remote_monitor), the ingest path never pauses for the user — so cfg.manual.remote_monitor=true will always raise RuntimeError after 30s, before the user can react. This directly contradicts the "harden" intent of the PR by breaking an
  existing workflow.
  6. Kill-fallback can cause total data loss. Because the monitor is a forked multiprocessing.Process, pkill -f in kill_remote_monitor() can SIGTERM the sampler child directly before the control_pipe stop handshake completes, leaving monitor_info=None. The null-JSON
  guard then deletes any partial monitor_output_ingest.json — so "escalate stuck samplers with kill()" can produce zero data instead of partial data

milindsrivastava1997:
**Hard violation #2 (unconditional `timeout=600` + force-kill escalation hitting `experiment_run_e2e.py`):** rather than threading `execution_mode`/`experiment_output_dir` scoping into the e2e.py call (which still leaves a "kill after N minutes" semantic that didn't exist before this PR), we removed the force-kill escalation from `wait_for_remote_monitor_to_finish()` entirely. It's back to its original pre-PR behavior — poll forever, no timeout, no kill — just with the polling now going through the new `is_remote_monitor_running()` helper. This resolves the issue at the source instead of papering over it with scoping, and doesn't introduce a new "wait forever" → "kill after 10 min" behavior change for callers outside this PR's scope. 

**Hard violation #1 (SIGTERM/SIGINT handler swallowing):** Fix: restore `SIG_DFL` for SIGTERM/SIGINT right after the ingest loop exits, so a later signal actually terminates the process again.

milindsrivastava1997:
Removed force kill of remote_monitor. Reverted back to behavior where we wait for remote_monitor to stop gracefully.

asap-tools/experiments/experiment_utils/services/remote_monitor_service.py
```diff
@@ -252,6 +268,183 @@ def start(
...
+            time.sleep(polling_interval)
+
+    def signal_ingest_monitor_stop(self, experiment_output_dir: str) -> None:
+        """Ask an ingest-mode ``remote_monitor.py`` to shut down gracefully."""
+        stop_file = os.path.join(
+            experiment_output_dir, constants.INGEST_MONITOR_STOP_FILE
+        )
+        self.provider.execute_command(
+            node_idx=self.node_offset,
+            cmd=f"touch {shlex.quote(stop_file)}",
```
milindsrivastava1997, line 453: ignore_errors=True on the stop-signal touch swallows SSH failures silently — the caller's finally block then polls with no timeout/SIGTERM fallback, so a transient SSH blip hangs the run indefinitely.

asap-tools/experiments/experiment_run_clickhouse.py
```diff
@@ -192,12 +202,127 @@ def _run_query_workload(
...
+        manual_mode=manual_remote_monitor,
+    )
+    if not manual_remote_monitor:
+        remote_monitor_service.wait_for_remote_monitor_start(
+            timeout=constants.REMOTE_MONITOR_START_TIMEOUT_SECONDS,
+            polling_interval=constants.REMOTE_MONITOR_START_POLL_INTERVAL_SECONDS,
+            execution_mode="ingest",
+            experiment_output_dir=baseline_output_dir,
+        )
+        time.sleep(1)
```
milindsrivastava1997: Bare time.sleep(1) here — every other timing value this PR touches got a named constant in constants.py, this one didn't.

asap-tools/experiments/experiment_utils/services/remote_monitor_service.py
```diff
@@ -252,6 +268,183 @@ def start(
...
+            cmd=cmd,
+            cmd_dir=None,
+            nohup=False,
+            popen=False,
+            ignore_errors=True,
+        )
+        assert isinstance(result, subprocess.CompletedProcess)
+        return bool(result.stdout.strip())
+
+    def wait_for_remote_monitor_start(
```
milindsrivastava1997: This hand-rolls a timeout/poll/raise loop that duplicates BaseService.wait_until_ready(), which this class inherits but never wires up.

asap-tools/experiments/experiment_utils/services/remote_monitor_service.py
```diff
@@ -252,6 +268,183 @@ def start(
...
+            ):
+                return
+            time.sleep(polling_interval)
+        mode_desc = execution_mode or "any"
+        raise RuntimeError(
+            f"remote_monitor.py ({mode_desc}) did not start within "
+            f"{timeout}s. Check remote_monitor.out under the experiment output dir."
+        )
+
+    def wait_for_remote_monitor_process_exit(
```
milindsrivastava1997, line 428: Three near-identical pgrep poll loops in this file now (382, 404, 477) with no shared helper — future timeout fixes will likely land in one and not the others.

milindsrivastava1997:
experiment_run_e2e.py: the stop() (~L270) and wait_for_remote_monitor_to_finish() (~L636) calls never pass the new execution_mode/experiment_output_dir, so the PR's stray-process fix only covers the new ClickHouse path, not the original Prometheus path.

asap-tools/experiments/experiment_utils/services/remote_monitor_service.py
milindsrivastava1997: addressed functionally, but not by calling BaseService.wait_until_ready(). We consolidated the polling into _wait_for_remote_monitor_state(). The base method is not a good fit because monitor matching requires per-call mode/directory filters and we
    also wait for process exit.

milindsrivastava1997 (approved):

Issue #474, opened 2026-06-29
fix(precompute): bulk single-timestamp ingest silently drops rows when wall-clock fallback closes the window mid-ingest


akanksha-akkihal:
## Summary

When ingesting a large one-shot batch where many rows share the same event-time (e.g. netflow data at 5M rows/s, all packed into one timestamp-second), the precompute engine **force-closes the tumbling window ~6s into ingest** via the wall-clock fallback. Rows that arrive after that point hit the "evicted pane" path and, because `late_data_policy` is hardcoded to `Drop`, are **silently discarded**.

The result is a uniform undercount across all group keys. Queries read a *closed but incomplete* window from the store.

## Observed behavior

5M rows/s netflow run (5M rows, all in window `[13:10:00, 13:10:01)`), sketchdb vs ClickHouse baseline:

| Query | baseline | sketchdb | error |
|-------|----------|----------|-------|
| `COUNT(pkt_len)` per srcip | 500,000 | 300,800 | **39.84%** low (all 10 srcips) |
| `SUM(pkt_len)` per srcip | ~453M | ~272M | **~39.8%** low |
| `COUNT(DISTINCT dstip)` per srcip | ~35,400 | ~25,700 | **~27%** low |

- `300,800 / 500,000 = 0.6016` → exactly ~60% of rows retained, ~40% dropped.
- Error is **uniform across every srcip** (rows are interleaved, so each key loses the same fraction).

## Root cause

All rows share one event-time, so the event-time watermark never advances. The window only closes via the wall-clock fallback in `worker.rs`, which fires after `window_size_ms + wall_clock_grace_period_ms` (default `1000 + 5000 = 6000` ms) of wall-clock time:

```rust
// asap-query-engine/src/precompute_engine/worker.rs:603
let window_size_ms = state.window_manager.window_size_ms();
for (&pane_start, &pane_birth_ms) in &state.pane_wall_clock_starts_ms {
    if now_ms.saturating_sub(pane_birth_ms) >= window_size_ms + grace_ms {
        let force_to = pane_start.saturating_add(window_size_ms);
        if force_to > effective_wm {
            effective_wm = force_to;
        }
    }
}
```

Ingesting 5M rows takes longer than 6s, so the window closes mid-ingest. Subsequent rows find their pane already evicted and are dropped:

```rust
// asap-query-engine/src/precompute_engine/worker.rs:430
match late_data_policy {
    LateDataPolicy::Drop => {
        debug!(
[...]

PR #538, opened 2026-07-27
test(precompute): add real-pipeline e2e regression test for #474

milindsrivastava1997:
## Summary
- Stacked on #537. Adds `JsonFileIngestConfig::batch_delay_ms` (default `0`, a no-op everywhere in production and in the existing `netflow_single_second` e2e test) — a real, explicit delay after each batch send, so an integration test can force ingest to span real wall-clock time deterministically instead of hoping incidental scheduling overhead is enough.
- New e2e test drives the real `JsonFileIngestSource → PrecomputeEngine → sink` pipeline with a bulk single-timestamp load spread across 10 batches, using #537's wall-clock override (unscaled — an earlier scaled version amplified ordinary scheduling jitter into spurious failures) and generous, seconds-scale margins so the periodic flush timer gets many real chances to land mid-ingest without depending on tight timing.
- Verified red against the pre-fix birth-time semantics (temporarily reverted the fix from #536, ran, confirmed rows were dropped mid-ingest, restored the fix) before counting it as a real regression test.
- Along the way, fixed `allowed_lateness_ms` in the new test's config: every `flush_all()` call unconditionally nudges the event-time watermark forward by 1ms (the "boundary advance" that lets an idle stream make progress), independent of the wall-clock fallback. With `allowed_lateness_ms=0` and multiple batches spaced across several flush ticks, that 1ms drift alone was enough to mark the next same-timestamp batch "late" via a wholly different code path, unrelated to #474. A generous, production-realistic value avoids that.

## Test plan
- [x] New e2e test passes; verified red against pre-fix code first.
- [x] Existing `e2e_netflow_single_second` and `e2e_precompute_equivalence` still pass.
- [x] Full `precompute_engine::` unit suite (81 tests) passes.
- [x] `cargo clippy -p query_engine_rust --lib --tests` clean.

### DIFF
--- a/asap-query-engine/src/precompute_engine/json_ingest.rs
+++ b/asap-query-engine/src/precompute_engine/json_ingest.rs
@@ -42,13 +42,10 @@ pub struct JsonFileIngestConfig {
     pub timestamp_col: String,
     pub timestamp_unit: TimestampUnit,
     pub batch_size: usize,
-    /// Test-support only: real delay after sending each batch, in
-    /// milliseconds. `0` (the only value any production config uses) is a
-    /// no-op — the reader sends batches as fast as it can, unchanged from
-    /// before this field existed. A nonzero value exists so integration
-    /// tests can force ingest to span real wall-clock time deterministically
-    /// (e.g. to give a periodic flush timer real chances to fire mid-ingest),
-    /// instead of hoping incidental scheduling overhead is enough.
+    /// Delay after sending each batch, in milliseconds. `0` (the default)
+    /// sends batches as fast as possible, unchanged from before this field
+    /// existed. A nonzero value throttles ingestion to a real, deterministic
+    /// pace.
     pub batch_delay_ms: u64,
 }
 
--- a/asap-query-engine/src/precompute_engine/json_ingest.rs
+++ b/asap-query-engine/src/precompute_engine/json_ingest.rs
@@ -42,6 +42,14 @@ pub struct JsonFileIngestConfig {
     pub timestamp_col: String,
     pub timestamp_unit: TimestampUnit,
     pub batch_size: usize,
+    /// Test-support only: real delay after sending each batch, in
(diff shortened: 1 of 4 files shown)


asap-query-engine/src/precompute_engine/json_ingest.rs
```diff
@@ -42,6 +42,14 @@ pub struct JsonFileIngestConfig {
     pub timestamp_col: String,
     pub timestamp_unit: TimestampUnit,
     pub batch_size: usize,
+    /// Test-support only: real delay after sending each batch, in
```
zzylol: If this is test only support, why "JsonFileIngestSource" is in this file? I would change the comment to a real json ingest support field.

asap-query-engine/tests/e2e_precompute_wall_clock_fallback_active_ingest.rs
```diff
@@ -0,0 +1,203 @@
+//! End-to-end regression test for issue #474: a bulk one-shot ingest whose
```
zzylol: Can we write it as a feature for json file ingestion test, not a regression test, or tied to a specific issue number in the comment/code?

zzylol (changes requested):

asap-query-engine/src/precompute_engine/json_ingest.rs
zzylol: If you have multiple various file ingestion, e.g., csv as well, better to group these similar configuration in the same struct, or a better organized definition for them

zzylol (changes requested):

asap-query-engine/src/precompute_engine/json_ingest.rs
milindsrivastava1997: Agree. Will skip it for this PR. Added #545 to track this

PR #542, opened 2026-07-28
fix(asap-tools): batch ClickHouse JSON-lines bulk load instead of a single oversized INSERT

### LINKED ISSUE #541: fix(asap-tools): batch ClickHouse JSON-lines bulk load instead of a single oversized INSERT
## Summary

Bulk-loading JSON-lines datasets into ClickHouse currently pipes the entire file through a single SSH → `docker exec` → one `INSERT … FORMAT JSONEachRow` command. That works for small fixtures but is unreliable for the multi-GB datasets used in the ClickHouse benchmark (`experiment_run_clickhouse.py`), where a single oversized INSERT gives no partial progress and no useful diagnostics when it fails.

We should load in bounded batches, wait for ClickHouse to actually be reachable first, and fail with an actionable message when input data is malformed.

akanksha-akkihal:
## Summary

Bulk-loading JSON-lines datasets into ClickHouse previously piped the entire file through a single `INSERT ... FORMAT JSONEachRow`, which is unreliable for the multi-GB datasets used in the ClickHouse benchmark. This PR loads in bounded batches, waits for ClickHouse to be reachable before loading, and fails with an actionable message when input data is malformed.

## Changes

- **Batched loader** (`experiment_utils/services/json/loader.py`): streams the
  file and inserts in bounded batches (default 100k rows) via
  `docker exec clickhouse-client`, instead of one giant INSERT.
- **Readiness check**: poll ClickHouse HTTP `/ping` until it returns `Ok.`
  before loading; on timeout, raise with the tail of the container logs.
- **Input validation**: each line is checked for valid JSON and embedded null
  bytes, so malformed input fails with the offending line number and a preview.
- **MV-safe init**: when `init_sql_file` is provided, it owns `DROP`/`CREATE`
  for its own objects — the standalone `DROP TABLE` no longer runs first and
  break schemas with dependent materialized views.
- Both the ClickBench and custom JSON-lines paths share the batched loader.

### DIFF
--- a/asap-tools/experiments/experiment_utils/services/clickhouse_service.py
+++ b/asap-tools/experiments/experiment_utils/services/clickhouse_service.py
@@ -459,7 +459,7 @@ class ClickHouseDataLoaderService(BaseService):
         dataset_label: str,
         batch_size: int = JSON_BATCH_SIZE,
     ) -> None:
-        """Load JSON-lines via batched HTTP INSERTs (safe for multi-GB files)."""
+        """Load JSON-lines in bounded batches via clickhouse-client."""
         local_script = os.path.join(self._ASSETS_DIR, self.JSON_LOADER_SCRIPT)
         remote_script = f"/tmp/json_loader_{os.getpid()}.py"
         self._rsync_to_remote(local_script, remote_script)
--- a/asap-tools/experiments/experiment_utils/services/clickhouse_service.py
+++ b/asap-tools/experiments/experiment_utils/services/clickhouse_service.py
@@ -482,7 +482,10 @@ class ClickHouseDataLoaderService(BaseService):
                 nohup=False,
                 popen=False,
             )
-            if isinstance(result, subprocess.CompletedProcess) and result.returncode != 0:
+            if (
+                isinstance(result, subprocess.CompletedProcess)
+                and result.returncode != 0
+            ):
                 detail = (result.stderr or result.stdout or "").strip()[:300]
                 raise RuntimeError(f"{dataset_label} data load failed: {detail}")
         finally:
@@ -514,9 +517,7 @@ class ClickHouseDataLoaderService(BaseService):
     ) -> None:
         """Stream a JSON-lines file (optionally gzipped) into ClickHouse."""
         print(f"Loading ClickBench data from {remote_data_file!r}...")
-        self._load_json_batched(
-            remote_data_file, url, table, max_rows, "ClickBench"
-        )
+        self._load_json_batched(remote_data_file, url, table, max_rows, "ClickBench")
 
     def _load_h2o(
         self, remote_data_file: str, url: str, batch_size: int, max_rows: int
@@ -568,6 +569,4 @@ class ClickHouseDataLoaderService(BaseService):
                 f"Unsupported file format for {remote_data_file!r}. "
                 "Use dataset_name='h2o' for CSV files."
             )
-        self._load_json_batched(
-            remote_data_file, url, table, max_rows, "Custom"
-        )
+        self._load_json_batched(remote_data_file, url, table, max_rows, "Custom")
--- a/asap-tools/experiments/experiment_utils/services/clickhouse_service.py
+++ b/asap-tools/experiments/experiment_utils/services/clickhouse_service.py
@@ -5,6 +5,7 @@ ClickHouse Docker service management and data loading for SQL experiment infrast
 import os
 import shlex
 import subprocess
+import time
 from typing import Optional
 from jinja2 import Template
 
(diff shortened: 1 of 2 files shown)


milindsrivastava1997:
Code review findings on this PR:

1. **`asap-tools/experiments/experiment_utils/services/clickhouse_service.py:317`** — The new ClickHouse HTTP readiness wait runs *after* the DROP TABLE / init-SQL execution, not before it, so it can't protect the very calls it was added to guard. `start()` runs `_exec_sql_file`/`DROP TABLE IF EXISTS` first and only calls `_ensure_clickhouse_http_ready(url)` afterward. If ClickHouse's HTTP port isn't up yet, the SQL exec calls (which curl the same url) raise immediately, before the readiness wait ever runs. Currently masked because the sole caller already waits via `ClickHouseService._wait_for_service_ready()` first.

2. **`asap-tools/experiments/experiment_utils/services/json/loader.py:106`** — `max_rows` no longer bounds how much of the file is read/validated: the early-exit check compares against `total`, which only advances inside `flush()` at batch boundaries (default `batch_size=100_000`), not per line read. With e.g. `--max-rows 1000` and the default batch size on a file with >100k lines, the loader reads/validates up to ~100k lines before truncating and stopping — a bad line anywhere in that window aborts the whole load even though only 1000 rows were requested. Sibling `h2o/loader.py` avoids this by checking the raw per-line index, not a post-flush counter.

3. **`asap-tools/experiments/experiment_utils/services/clickhouse_service.py:313`** (plausible) — The unconditional `DROP TABLE IF EXISTS` that used to run before every load is now only issued inside the `BUILTIN_DDL_FILES` branch, so `dataset_name="custom"` with no `init_sql_file` no longer drops/resets the table between runs — repeated runs would append instead of starting clean.

4. **`asap-tools/experiments/experiment_utils/services/clickhouse_service.py:413`** — `_ensure_clickhouse_http_ready` reimplements the same poll-`/ping`-with-retries loop that `DockerServiceBase._wait_for_service_ready` / `ClickHouseService.is_healthy()` already provide, instead of reusing that mechanism. Not a bug, but risks the two copies drifting out of sync.

5. **`asap-tools/experiments/experiment_utils/services/clickhouse_service.py:439`** — The timeout-path log-fetch hardcodes the container name `"clickhouse-server"` instead of using `ClickHouseService.CONTAINER_NAME`, unlike `_load_json_batched` nearby which does it correctly.

Will follow up with fixes for these.

asap-tools/experiments/experiment_utils/services/clickhouse_service.py
```diff
@@ -403,6 +426,68 @@ def _exec_sql_file(self, remote_sql_file: str, url: str) -> None:
...
+                "--batch-size {} --max-rows {} --container {}"
+            ).format(
+                shlex.quote(remote_script),
+                shlex.quote(remote_data_file),
+                shlex.quote(table),
+                batch_size,
+                max_rows,
+                shlex.quote(ClickHouseService.CONTAINER_NAME),
+            )
+            result = self.provider.execute_command(
```
milindsrivastava1997, line 479: Missing `ignore_errors=True` — `execute_command` will raise `CalledProcessError` on non-zero exit before the `CompletedProcess`/returncode check below ever runs, so this error path is dead code.

asap-tools/experiments/experiment_utils/services/clickhouse_service.py
```diff
@@ -286,14 +303,17 @@ def start(
             )
 
         url = f"http://localhost:{self.clickhouse_http_port}/"
-
-        print(f"Dropping table {table!r}...")
-        self._exec_sql(f"DROP TABLE IF EXISTS {table}", url)
+        self._ensure_clickhouse_http_ready()
 
         if init_sql_file is not None:
```
milindsrivastava1997, line 308: DROP TABLE is skipped when `init_sql_file` is set. `quantile_demo/init.sql` is just `CREATE TABLE IF NOT EXISTS`, so a second run silently keeps stale rows.

asap-tools/experiments/experiment_utils/services/clickhouse_service.py
```diff
@@ -403,6 +426,68 @@ def _exec_sql_file(self, remote_sql_file: str, url: str) -> None:
...
+                log_tail = (logs.stdout or logs.stderr or "").strip()[-500:]
+            raise RuntimeError(
+                "ClickHouse HTTP is not ready before data load. "
+                f"Check clickhouse_logs/ on the experiment node. Recent logs: {log_tail}"
+            ) from exc
+
+    def _load_json_batched(
+        self,
+        remote_data_file: str,
+        url: str,
```
milindsrivastava1997: `url` param is unused (leftover from the old curl/HTTP path).

asap-tools/experiments/experiment_utils/services/clickhouse_service.py
```diff
@@ -403,6 +426,68 @@ def _exec_sql_file(self, remote_sql_file: str, url: str) -> None:
         for stmt in (s.strip() for s in result.stdout.split(";") if s.strip()):
             self._exec_sql(stmt, url)
 
+    def _ensure_clickhouse_http_ready(self, max_retries: int = 30) -> None:
```
milindsrivastava1997: `max_retries=30` here means 60s (2s poll), but `DockerServiceBase._wait_for_service_ready`'s `max_retries=30` means 150s (5s poll) — same name, different semantics. Also duplicates that readiness-wait logic rather than extending it.

milindsrivastava1997 (approved):

PR #562, opened 2026-08-20
fix(query-engine): sliding windows validate range-query step against slide interval

### LINKED ISSUE #557: Support sliding window candidates in the query engine (execution)
Sub-issue of #554.

Scope: actually serving queries correctly against sliding-window aggregations — the execution-side work, separate from generating valid candidates (#554's other sub-issue, #556).

Concretely:
- Relax `capability_matching.rs`'s `window_compatible()` Sliding rule beyond exact-match (`data_range_ms == window_size_ms`) to allow `data_range_ms` a multiple of `window_size_ms`, mirroring Tumbling's existing rule.
- Implement merge (mergeable sketch types) / subtract (subtractable types) across multiple stored sliding-window buckets in `simple_engine`'s query-serving path (`create_store_query_plan`/`execute_and_merge_store_queries` in `simple_engine/mod.rs`, which today unconditionally does a single exact-match lookup for Sliding and explicitly skips merging).
- Update `planner/cleanup.rs`'s `WindowType::Sliding` branch to be window-count-aware instead of hardcoding `read_count_threshold = 1`.
- Resolve the epoch/slide-interval alignment question for exact-match lookups (`align_end_timestamp_promql` only aligns to the scrape interval, not the aggregation's `slide_interval_ms`).
- Owns the end-to-end correctness test (plan → run → query a sliding-window aggregation → assert correct results) — #555 explicitly defers this test here.

Needs its own design session — not scoped further here. See #554 and the design docs linked there for context.

milindsrivastava1997:
Part of #557. First of a 4-PR stack (design doc: `.design_docs/sliding-window-query-engine-design.md`).

## What

`validate_range_query_params` rejected any range-query `step` that wasn't a multiple of `window_size_ms`, regardless of window type. For Sliding this is stricter than necessary — see the doc's §5 for the full reasoning. Adds a Sliding-only branch checking `step % slide_interval_ms == 0` instead; Tumbling's existing check is untouched.

## Why this is safe to merge on its own

`should_use_sliding_window()` is still hardcoded `false` — the planner cannot emit a `Sliding` config today, so nothing in production can reach the new branch. Tumbling's behavior is byte-for-byte unchanged.

## Stack

This is PR A. PR B (execution unification), PR C (`cleanup.rs`), and PR D (`capability_matching.rs` relaxation — the one that actually activates multi-window Sliding serving) will stack on top of this one, each as its own reviewable PR.

## Testing

7 new unit tests (`validate_range_query_params_tests`) covering both window types plus the pre-existing `start < end` / `step > 0` checks. Full workspace `cargo test`, `cargo clippy -D warnings`, `cargo fmt --check` all pass.

### DIFF
--- a/.design_docs/sliding-window-query-engine-design.md
+++ b/.design_docs/sliding-window-query-engine-design.md
@@ -98,13 +98,14 @@ extended to loop. Net: less code than today.
 ### 4. Epoch/slide-interval alignment
 
 Worker buckets sit on the `slide_interval_ms` grid, anchored at Unix epoch 0
-(`precompute_engine/window_manager.rs`). Two things need alignment, and
-they're different in kind:
+(`precompute_engine/window_manager.rs`). The fix differs between instant and
+range queries, because §5 gives range queries a way to guarantee alignment
+by construction instead of re-deriving it defensively on every step.
 
-- **The fetch bounds** (`[end - k*W, end)`) don't need to be grid-aligned —
-  `query_precomputed_output` is a plain numeric range scan, not a grid
-  lookup, so a generously-bounded span is fine as long as it covers
-  everything needed.
+- **The fetch bounds** (`[end - k*W, end)`), in both cases, don't need to be
+  grid-aligned — `query_precomputed_output` is a plain numeric range scan,
+  not a grid lookup, so a generously-bounded span is fine as long as it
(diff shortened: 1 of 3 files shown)


milindsrivastava1997:
Subsumed by other PRs like #640

Issue #567, opened 2026-08-21
asap-query-engine: convert binary PromQL instant queries from DataFusion to native execution


milindsrivastava1997:
## Context

`asap-query-engine`'s `SimpleEngine` currently serves PromQL binary-arithmetic
instant queries (vector-vector and vector-scalar) through a dedicated
DataFusion path (`handle_binary_expr_promql` → `build_arm_logical_plan` →
`build_binary_vector_plan`/`build_scalar_plan` → `execute_logical_plan`), while
every other live query shape (plain vector-selector instant queries, both
PromQL range-query shapes, all of SQL, all of Elastic DSL) is served by the
native/direct path (`execute_query_pipeline` → `execute_and_merge_store_queries`
→ `Store::query_precomputed_output[_exact]` + hand-written merge).

This is the only place DataFusion is reachable in production at all. Having
one query shape run through a structurally different engine means the two
implementations of "fetch precomputed outputs and merge them" have to be kept
behaviorally equivalent by hand — and they already aren't (see design
decisions below).

Full audit this issue was scoped from: `.design_docs/query-engine-fetch-merge-audit.md`.

## Design decisions (resolved via grilling session, 2026-08-21)

- **Sliding-bucket merge bug is real and must be fixed first, its own stage.**
  `execute_and_merge_store_queries` (`simple_engine/mod.rs:551-567`) takes the
  first bucket per key for Sliding-window queries and discards the rest
  (warn-only). `SummaryMergeMultipleExec` (DataFusion) merges all of them
  correctly. Binary-expr queries are currently the one shape immune to this,
  because they're the one shape still on DataFusion — cutting them over
  as-is would silently inherit the bug. DataFusion's merge-all behavior is
  the correct one; native gets fixed to match, not the other way round.
  Expected-count-per-key is hardcoded to `1` for now (`ponytail:` comment
  pointing at #554, which will make >1 legitimate — don't block on it).
  **Done: #570.**

- **Label-order join "bug" investigated, no live bug found — dropped.**
  Originally scoped as its own stage: the vector-vector join
  (`handle_binary_expr_range_promql`) keys a `HashMap` by raw
  `KeyByLabelValues` (positional `Vec<String>`), and the concern was that two
  metrics sharing a label set could declare it in different order, silently
  breaking the join. Turns out this can't happen: `KeyByLabelNames::new()`
  (`promql_utilities/data_model/key_by_label_names.rs:10-18`)
[...]

PR #582, opened 2026-08-23
fix(query-engine): range queries expand keys_query and merge same-timestamp buckets

### LINKED ISSUE #580: Range query pipeline drops keys_query and hardcodes is_exact_query=false
`execute_range_query_pipeline`, used by PromQL range queries (both plain and binary-expr, since the binary-expr range path calls this per arm), has two problems:

1. It never reads `keys_query`. Dual-population metrics silently lose the keys side over a range, unlike the instant-query path, which handles both keys and values.
2. It unconditionally forces `is_exact_query = false` regardless of the aggregation's actual `WindowType` (`promql.rs:1144`). Sliding-window aggregations are merged incorrectly for every range query as a result.

milindsrivastava1997:
## Summary
- `execute_range_query_pipeline` never read `keys_query`, so dual-population metrics (separate value/key aggregations) returned an empty result over a range instead of the expanded key set.
- The per-step bucket lookup collapsed same-start-timestamp buckets into a `HashMap<u64, _>`, silently dropping all but one when a Sliding aggregation legitimately returns more than one bucket per window (#567/#570).
- Both fixes mirror the fetch/merge/expand pattern already used by the instant-query path (`execute_and_merge_store_queries` / `collect_results_separate_keys`).

Fixes #580.

## Test plan
- [x] Added `native_range_query_tests.rs`: RED before the fix, GREEN after — dual-population single-step, dual-population multi-step, sliding 2-bucket collision, sliding 3-bucket collision, sliding single-bucket regression guard.
- [x] `cargo test -p query_engine_rust --lib` — 552 passed, 0 failed.
- [x] `cargo clippy -p query_engine_rust --lib --tests` — clean.

### DIFF
--- a/asap-query-engine/src/engines/simple_engine/mod.rs
+++ b/asap-query-engine/src/engines/simple_engine/mod.rs
@@ -1456,6 +1456,22 @@ impl SimpleEngine {
             all_data.values().map(|v| v.len()).sum::<usize>()
         );
 
+        // Dual-population metrics (separate key/value aggregations) need the
+        // keys side merged once up front, then expanded per value group below
+        // — mirrors execute_and_merge_store_queries/collect_results_separate_keys
+        // (#580: this pipeline previously never read keys_query at all).
+        let merged_keys: Option<MergedOutputsMap> =
+            if let Some(keys_params) = &context.base.store_plan.keys_query {
+                let keys_map = self.execute_store_query(keys_params)?;
+                Some(self.merge_precomputed_outputs(
+                    &keys_map,
+                    true,
+                    context.base.agg_info.aggregation_type_for_key,
+                ))
+            } else {
+                None
+            };
+
         let mut results: HashMap<KeyByLabelValues, RangeVectorElement> = HashMap::new();
 
         // Determine accumulator type for merger selection
@@ -1467,6 +1483,8 @@ impl SimpleEngine {
         let end_ms = context.range_params.end;
         let buckets_per_step = context.buckets_per_step;
         let lookback_bucket_count = context.lookback_bucket_count;
+        let tumbling_window_ms = context.tumbling_window_ms;
+        let lookback_ms = (lookback_bucket_count as u64) * tumbling_window_ms;
 
         let window_mode = if buckets_per_step <= lookback_bucket_count {
             "sliding (slide <= size)"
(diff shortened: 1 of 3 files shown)


milindsrivastava1997:
Code review findings on `execute_range_query_pipeline` (`asap-query-engine/src/engines/simple_engine/mod.rs`):

1. **mod.rs:1512** — Range key expansion silently skips a value group (`continue`, no log/error) when `get_keys()` returns `None` or the group key is missing from `merged_keys`, while the mirrored instant-query path returns an `Err` for the same condition. A dual-population metric using `DeltaSetAggregator` whose key set ever shrank (`get_keys()` returns `None` per `delta_set_aggregator_accumulator.rs:251-258`) causes this path to silently return an empty-but-Ok result instead of erroring, so `handle_range_query_promql`'s Prometheus fallback (which only triggers on `Err`) never kicks in — callers get a wrongly-empty successful response.

2. **mod.rs:1463** — The keys aggregation is fetched and merged once up front and reused unchanged for every output timestep in the range loop, even though key membership can legitimately change over the queried interval. For a dual-population range query where labels are added/removed partway through the range, every timestep gets the same key expansion from the final snapshot — producing phantom samples for keys that didn't exist yet early on, or missing series for keys gone by the range's end.

3. **mod.rs:1463** — The new keys-fetch-and-merge block duplicates equivalent logic already in `execute_and_merge_store_queries` (mod.rs:596-625) instead of sharing a helper. The two copies have already diverged (instant path passes the real `do_merge` flag and logs latency; range path hardcodes `true` and drops logging), so a future fix to one merge path is likely to be applied to only one copy.

milindsrivastava1997:
Follow-up finding on the latest push (`7b9f084`), in `execute_range_query_pipeline`:

**`keys_query`'s fetch window is never widened for range queries, only `values_query` is** (`asap-query-engine/src/engines/simple_engine/mod.rs:1491`, `promql.rs:583`). `finish_range_context` widens `values_query` to `[start-lookback, end]` but clones `keys_query` unchanged, and `keys_query`'s window is computed from a single instant (`create_keys_query_params`, using only `end_timestamp`):

- `SetAggregator` ("latest window only") → keys window is just `[end-window_size, end]`, the *last* window of the range.
- `DeltaSetAggregator` ("all keys since start") → keys window is `[0, end]`, all of history.

This causes two different failure modes:

1. **Silent data loss (SetAggregator):** the range loop iterates over the keys present in `merged_keys`, not over `all_data`. A label with real values earlier in the range but absent from the final window (e.g. a pod that scaled down) never gets iterated at all — its whole history silently disappears from the output, no error.

2. **Whole-query failure (DeltaSetAggregator):** `keys_query`'s `[0, end]` window picks up keys from arbitrarily far in the past — including ones whose only data point is well before the queried range and outside the (still-bounded) widened `values_query` window. That key is in `merged_keys` but missing from `all_data`, and the new hard-error semantics (`ok_or_else(...)?` at mod.rs:1541) fail the *entire* range response over one stale key, even though every other key in the query has good data.

None of the new tests in `native_range_query_tests.rs` exercise mismatched key/value windows (all use 1:1 matching timestamps), so this path is untested.

milindsrivastava1997:
Correction on the DeltaSetAggregator part of the finding above.

\`merge_accumulators\` in \`delta_set_aggregator_accumulator.rs\` does correctly reconcile add/remove deltas across \`[0, T]\` — a key that's added then removed cancels out. So it's not "picks up every key that ever existed" as I said.

The real trigger is simpler: \`get_keys()\` (delta_set_aggregator_accumulator.rs:251-258) returns \`None\` for the *entire* accumulator if \`removed\` is non-empty post-merge, instead of returning \`added.difference(&removed)\`. Since \`removed\` is non-empty as soon as any key has *ever* been removed across the metric's full history, this fires on ordinary label churn, not just stale/ancient keys. Pre-existing (since #342), not introduced by this PR — but #582's new hard-fail on missing keys (mod.rs:1547) turns this common case into a full range-query failure instead of the previous silent skip.

PR #587, opened 2026-08-24
fix(query-engine): expand self-keyed accumulators in range queries

### LINKED ISSUE #584: Range queries drop self-keyed accumulator expansion (top-k) for single-population metrics
`execute_range_query_pipeline`'s single-population branch (no separate `keys_query`) uses each value group's own `group_key` directly and never calls `get_keys()` on the value accumulator itself. The instant path (`collect_results_same_aggregation`) does call it, which is how self-keyed accumulators like `CountMinSketchWithHeap` (top-k) expand into multiple output keys without a separate keys aggregation. A range query over such a metric, grouped by `None` at the store level, returns empty instead of the expanded top-k keys.

This predates #580/#582 — the original code also unconditionally skipped `None`-keyed groups in this branch — so it's a distinct, pre-existing gap, not a regression from that fix.

milindsrivastava1997:
## Summary
- `execute_range_query_pipeline`'s single-population branch never called `get_keys()` on the merged value accumulator, so self-keyed accumulators like `CountMinSketchWithHeap` (top-k) returned empty over a range instead of expanding into their top-k keys — the instant path already does this via `collect_results_same_aggregation`.
- Fix: every group now tries `merged.get_keys()` first (mirroring the instant path exactly) and only falls back to the precomputed key list when it returns `None`. This also covers self-keyed accumulators stored under a non-`None` outer key (real Arroyo/worker.rs ingestion always wraps `Some(key)`, even for empty grouping), not just the `None`-keyed case from the original report.

Fixes #584.

## Test plan
- [x] `cargo test -p query_engine_rust --lib native_range_query_tests` — 7 passed, including two new regression tests for #584 (`range_query_self_keyed_topk_expands_without_keys_query`, `range_query_self_keyed_topk_expands_with_non_none_outer_key`)
- [x] Full lib suite: `cargo test -p query_engine_rust --lib` — 554 passed, 0 failed
- [x] New tests independently verified red (against pre-fix code) / green (against fixed code)

### DIFF
--- a/asap-query-engine/src/engines/simple_engine/mod.rs
+++ b/asap-query-engine/src/engines/simple_engine/mod.rs
@@ -1525,13 +1525,16 @@ impl SimpleEngine {
             window_mode
         );
 
-        // Resolve, for every value group, which output label-keys it serves:
-        // its own key for single-population metrics, or every key the merged
-        // keys aggregation expands it to for dual-population metrics. Mirrors
-        // collect_results_separate_keys exactly, including its error
-        // semantics — an unresolvable key set fails the whole range query
-        // (so callers fall back to Prometheus) instead of silently returning
-        // a partial result. See #582 review.
+        // Resolve, for every value group, a fallback key list — used only if
+        // the merged value accumulator itself doesn't self-key (see below).
+        //  - dual-population metrics: every key the merged keys aggregation
+        //    expands it to. Mirrors collect_results_separate_keys exactly,
+        //    including its error semantics — an unresolvable key set fails
+        //    the whole range query (so callers fall back to Prometheus)
+        //    instead of silently returning a partial result. See #582
+        //    review.
+        //  - single-population metrics: just its own store-level group key
+        //    (0 or 1 keys).
         let groups: Vec<(
             &Vec<crate::stores::TimestampedBucket>,
             Vec<KeyByLabelValues>,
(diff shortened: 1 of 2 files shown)


milindsrivastava1997:
`execute_range_query_pipeline`'s `merged.get_keys().unwrap_or_else(|| fallback_keys.clone())` (mod.rs:1612) applies unconditionally to every group, including dual-population groups where `keys_query.is_some()`. That diverges from the instant path (`collect_all_results`), which branches globally: when a separate keys aggregation exists, `collect_results_separate_keys` is used and never consults the value accumulator's own `get_keys()`.

This is reachable in production: `count_topk_capability_fallback_pairs_heap_with_key_agg` (sql.rs:1443) shows capability-matching pairs `CountMinSketchWithHeap` (value) with `DeltaSetAggregator` (keys) — a real dual-population, self-keyed-value config. With this change, each window step now uses the heap's own capacity-limited per-window top-k instead of the keys-aggregation's stable, full key set, so a key with low count in a given window can silently drop out of that step (or the set can flap between steps) even though it's present in the instant-query result and every other window.

The new tests (`range_query_self_keyed_topk_expands_without_keys_query`, `..._with_non_none_outer_key`) only cover the single-population case, so this gap isn't caught by CI.

Suggest gating the `merged.get_keys()` fallback on whether the group came from the single-population branch, so dual-population groups always use the keys-aggregation-derived `fallback_keys` — matching `collect_results_separate_keys`.

PR #595, opened 2026-08-24
fix(query-engine): range queries expand keys_query per output step

### LINKED ISSUE #583: Range query key expansion uses one snapshot instead of per-step keys
`execute_range_query_pipeline`'s dual-population handling (added in #582, fixing #580) fetches and merges `keys_query` once, anchored at the range's `end`, and reuses that single snapshot for every output timestamp in the range. If the key set changes partway through the queried interval — a label combination first appears, or stops appearing — every step still uses the final snapshot: keys get phantom samples before they existed, or keep appearing after they stopped, instead of reflecting the key set as it actually was at each step's time.

Please also check other parts of the codebase for the same class of bug: any code path that fetches/merges a value or key set once and then reuses it across multiple output points in a time series, where the underlying set can legitimately change over that span. `handle_binary_expr_range_promql`'s per-arm range context (`build_arm_range_context`, which calls the same `finish_range_context`/keys-snapshot logic per arm) is one likely candidate worth checking specifically.

## Related: keys_query window scoping is instant-anchored, not range-aware

`finish_range_context` widens `values_query` to `[start-lookback, end]` for a range query, but clones `keys_query` unchanged — its window is still computed from a single instant (`create_keys_query_params`, using only `end_timestamp`), which is wrong for both key aggregation types, in opposite directions:

- **`SetAggregator`** ("latest window only"): keys window is `[end-window_size, end]`. Since the range loop iterates `merged_keys` (not `all_data`), a label with real values earlier in the range but absent from that final window's key snapshot never gets iterated at all.
- **`DeltaSetAggregator`** ("all keys since start"): keys window is `[0, end]`. This is actually *necessary* because this aggregator is a "delta" version of `SetAggregator`. It requires replaying every delta bucket back to the initial full-snapshot window

Fixing this properly likely means fetching/merging keys per output step, scoped consistently with each step's own window, rather than once at the range's `end` — which would fix both this scoping mismatch and the snapshot-staleness problem above as the same fix.

milindsrivastava1997:
## Summary
- Closes #583 
- `execute_range_query_pipeline` fetched/merged `keys_query` once (anchored at the range's `end`) and reused that single snapshot for every output timestamp — a key that appeared/disappeared mid-range got phantom or missing samples, including through the binary-expr arm path
- `keys_query`'s window was instant-anchored, not widened for the range the way `values_query` already is — wrong span for `SetAggregator` specifically
- Fixes both by fetching keys raw once (mirroring the existing values-side fetch) and merging them per output step, scoped to that step's own window, instead of once globally
- A group with keys data but no value data anywhere now gets skipped with a warning instead of hard-failing the entire range query
- Extracted `widen_query_window` to dedup the values/keys window-widening formula

### DIFF
--- a/asap-query-engine/src/engines/simple_engine/mod.rs
+++ b/asap-query-engine/src/engines/simple_engine/mod.rs
@@ -617,9 +617,11 @@ impl SimpleEngine {
     }
 
     /// Fetches and merges the keys side of a dual-population query plan, if
-    /// present. Shared by `execute_and_merge_store_queries` (instant) and
-    /// `execute_range_query_pipeline` (range) — both build a `StoreQueryPlan`
-    /// that may carry a separate `keys_query` and need it merged the same way.
+    /// present. Used by `execute_and_merge_store_queries` (instant) only —
+    /// `execute_range_query_pipeline` (range) fetches keys raw via
+    /// `execute_store_query` and merges them per output step instead (#583),
+    /// since a single merged snapshot can't answer "what did the key set
+    /// look like at an earlier timestamp."
     fn fetch_and_merge_keys(
         &self,
         keys_query: &Option<StoreQueryParams>,
@@ -1476,6 +1478,45 @@ impl SimpleEngine {
     //     Some((output_labels, QueryResult::matrix(range_elements)))
     // }
 
+    /// Builds a lookup from bucket start-timestamp to every bucket sharing
+    /// that start. A Sliding aggregation can legitimately return more than
+    /// one bucket per start timestamp (#567/#570) — every one of them must
+    /// be merged, not just the last one collected here. Used identically by
+    /// `execute_range_query_pipeline` for both the value side and (#583)
+    /// the keys side.
+    fn build_bucket_map(
+        buckets: &[crate::stores::TimestampedBucket],
+    ) -> HashMap<u64, Vec<&dyn AggregateCore>> {
+        let mut bucket_map: HashMap<u64, Vec<&dyn AggregateCore>> = HashMap::new();
+        for ((start, _), bucket) in buckets {
+            bucket_map.entry(*start).or_default().push(bucket.as_ref());
+        }
+        bucket_map
+    }
+
+    /// Collects every bucket in `bucket_map` whose start falls in
+    /// `[window_start, window_end)`, stepping by `step_increment`. Missing
+    /// buckets at a given start are skipped (partial data is okay). Used
+    /// identically by `execute_range_query_pipeline` for both the value
+    /// side and (#583) the keys side — the only difference between the two
+    /// call sites is which map/lookback/bucket-width they pass in.
+    fn scan_window(
+        bucket_map: &HashMap<u64, Vec<&dyn AggregateCore>>,
+        window_start: u64,
+        window_end: u64,
+        step_increment: u64,
+    ) -> Vec<Box<dyn AggregateCore>> {
+        let mut window_buckets: Vec<Box<dyn AggregateCore>> = Vec::new();
+        let mut t = window_start;
+        while t < window_end {
+            if let Some(buckets) = bucket_map.get(&t) {
+                window_buckets.extend(buckets.iter().map(|b| b.clone_boxed_core()));
+            }
+            t += step_increment;
+        }
+        window_buckets
+    }
+
     /// Execute the range query pipeline
     fn execute_range_query_pipeline(
         &self,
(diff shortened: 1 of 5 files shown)


milindsrivastava1997:
## Code review findings

### 1. `asap-query-engine/src/engines/simple_engine/mod.rs:1699` — Keys window scan is O(epoch_time/step) per step/group [CONFIRMED]
For a DeltaSetAggregator keys aggregation, the per-step key window scan iterates from timestamp 0 up to `current_time` in increments of `keys_tumbling_window_ms` — an O(current_time / step) loop run once per group per output timestamp.

**Failure scenario:** `create_keys_query_params` sets a DeltaSetAggregator's instant keys window to `[0, end_timestamp]`, so `keys_lookback_ms == end_ms`; in the per-step loop, `keys_window_start = current_time.saturating_sub(keys_lookback_ms)` is 0 for every `current_time <= end_ms`, so `scan_window` walks from 0 to `current_time` (real epoch-ms, ~1.7e12 today) stepping by `keys_tumbling_window_ms` (e.g. 30s–1h). At real deployment timestamp scale this is hundreds of thousands to tens of millions of HashMap probes per (group, output-step) pair — a severe latency blowup or effective hang for any range query over a DeltaSetAggregator-keyed dual-population metric.

### 2. `asap-query-engine/src/engines/simple_engine/mod.rs:1699` — Keys scan steps by window_size, not slide_interval, for Sliding [CONFIRMED]
`scan_window` steps the keys bucket_map by `keys_tumbling_window_ms` (window_size_ms), but stored bucket start-timestamps for a Sliding-type key aggregation are multiples of the separate `slide_interval_ms` field.

**Failure scenario:** A SetAggregator keys aggregation configured with `window_type=Sliding, window_size_ms=900_000, slide_interval_ms=30_000` stores buckets at 30s-aligned starts, but the per-step scan only checks timestamps 900_000ms apart — landing on a real bucket only 1-in-30 times. Most output steps silently resolve zero expansion keys and get skipped, dropping most of the range-query output. Design doc (`docs/583-range-keys-per-step-design.md`, Q8) confirms this configuration is untested.

### 3. `asap-query-engine/src/engines/simple_engine/promql.rs:586` — keys_tumbling_window_ms never validated nonzero [PLAUSIBLE]
`validate_range_query_params` only validates the value side's `tumbling_window_ms`; `keys_tumbling_window_ms` is never validated to be nonzero.

**Failure scenario:** If a key aggregation's `AggregationConfig.window_size_ms` is ever 0, `scan_window`'s `t += step_increment` becomes `t += 0`, looping forever and hanging the range query — with no equivalent protection to the value side's validated `tumbling_window_ms`.

### 4. `asap-query-engine/src/engines/simple_engine/mod.rs:1631` — Range path skips missing value data; instant hard-fails [CONFIRMED]
The range-query path now `warn!`+skips a group with keys data but no value data instead of failing; its comment claiming this mirrors the instant path is false — the instant path still hard-fails via `.ok_or_else` for the identical condition.

**Failure scenario:** The same misconfigured metric (a keys_query group with no matching value data) produces a hard, diagnosable error for an instant query but a silent, successful-looking range-query response with fewer series than expected and only a warn! log — a caller can't detect the discrepancy without reading server logs.

### 5. `asap-query-engine/src/engines/simple_engine/mod.rs:1712` — Unresolved remove-without-add now silently blanks output [PLAUSIBLE]
A DeltaSetAggregatorAccumulator with an unresolved 'remove' and no matching 'add' in its per-step merge window legitimately returns `None` from `get_keys()`; the new code folds that into an empty Vec and silently skips the step (debug! only), where the old code hard-failed the whole query.

**Failure scenario:** The DeltaSetAggregator keys window always starts at 0, implicitly assuming the store retains every 'add' bucket back to true t=0. If store retention is shorter than the query's full history (any real eviction policy), an old 'add' bucket can be evicted while a later 'remove' for the same key survives, producing an unresolved removal — silently and permanently blanking that group's output from that point on, with no error surfaced above debug level.

### 6. `asap-query-engine/src/engines/simple_engine/promql.rs:614` — Config race drops whole range query via `?` on key lookup [PLAUSIBLE]
`keys_tumbling_window_ms` is resolved via a second, independent `streaming_config.read()`/`get_aggregation_config` call inside `finish_range_context`, using `?` on the Option — a failure here discards the already-successfully-built value-side context too.

**Failure scenario:** `base_context.store_plan` is built from one streaming_config snapshot; if the config is swapped between that point and the second `get_aggregation_config` lookup (e.g. the planner applier rotates configs mid-request), `finish_range_context` returns `None` via `?`, throwing away the value side's already-resolved context and returning an empty/failed range query instead of degrading gracefully.

### 7. `asap-query-engine/src/engines/simple_engine/mod.rs:1690` — Value-window and key-window merge blocks duplicated inline [CONFIRMED]
The per-step value-window scan-and-merge and the new per-step key-window scan-and-merge are two nearly-identical ~15-line blocks hand-duplicated inline instead of factored into a shared helper.

**Failure scenario:** Any future fix to the windowing logic (e.g. the slide_interval_ms bug above) has to be applied twice, in two slightly-differently-shaped blocks, risking the two copies drifting further out of sync the way keys-vs-values behavior already diverged once in this PR.

### 8. `asap-query-engine/src/engines/simple_engine/mod.rs:1487` — Bucket map order not guaranteed chronological [PLAUSIBLE]
`build_bucket_map` pushes same-start-timestamp buckets in store-return order with no explicit chronological sort; this map is now reused for the keys side where `NaiveMerger`'s sequential left-fold is order-sensitive for `DeltaSetAggregatorAccumulator`.

**Failure scenario:** The PR's own new `window_merger.rs` test proves merging DeltaSetAggregator toggle-buckets out of chronological order yields the wrong key-present/absent state. If a DeltaSetAggregator key aggregation is ever paired with Sliding window_type (multiple buckets per start timestamp) and store return order for same-start buckets isn't guaranteed chronological, the per-step key merge could silently compute the wrong key set. Not confirmed reachable in current configs.

---

PR #606, opened 2026-08-25
fix(planner): restrict DeltaSetAggregator to tumbling windows

### LINKED ISSUE #588: DeltaSetAggregator must be restricted to tumbling windows
DeltaSetAggregator only tracks added/removed keys since the last window, so it's only correct for non-overlapping (tumbling) windows. Nothing currently enforces this: the planner (`agg_config.rs`) copies whatever `window_type` the surrounding query config has when emitting a DeltaSetAggregator config, and the precompute engine's `DeltaSetAggregatorAccumulator`/pane-merge path is window-type-agnostic. A sliding-window DeltaSetAggregator config can be planned and executed today, silently producing incorrect merged add/remove sets across overlapping panes.

Consequence for the query engine: within a single query, key-cardinality accumulators may need different merge strategies per window type — DeltaSetAggregator panes must be read/merged as tumbling windows, while other accumulators (e.g. SetAggregator, CountMinSketch, HydraKLL) tied to the same query may legitimately use sliding windows and need the sliding merge path.

Needs:
- Planner-side validation/gate rejecting or refusing to select DeltaSetAggregator when window_type is Sliding.
- Query-engine read path to handle mixed window types within one query (tumbling merge for DeltaSetAggregator, sliding merge for co-located accumulators).

milindsrivastava1997:
## Summary
- DeltaSetAggregator only tracks added/removed keys since the last window, so it's only correct for non-overlapping (tumbling) windows. Nothing enforced this, so a Sliding DeltaSetAggregator config could be planned/selected and silently produce wrong merged add/remove sets.
- `agg_config.rs`: force the DeltaSetAggregator companion config to Tumbling regardless of the surrounding query's window_type, sized to the sibling's `slide_interval_ms`.
- `candidate_gen.rs`: the (not-yet-wired-in) cost optimizer no longer enumerates Sliding candidates for DeltaSetAggregator.
- `capability_matching.rs`: new `key_agg_window_valid()` gate, wired into `window_compatible()` and the key-agg pairing lookup (which previously bypassed window checking entirely) — the actual belt-and-suspenders fix that stops a bad config from ever being selected to serve a query.

Closes #588.

## Test plan
- [x] TDD: red tests written per fix site, confirmed failing, then made to pass
- [x] `cargo test -p asap_planner -p query_engine_rust -p asap_types --lib` — 695 tests pass, 0 failures

### DIFF
--- a/asap-common/dependencies/rs/asap_types/src/capability_matching.rs
+++ b/asap-common/dependencies/rs/asap_types/src/capability_matching.rs
@@ -270,19 +270,23 @@ pub fn find_compatible_aggregation(
     // separate key aggregation just like any other multi-population value type.
     let key_agg: &AggregationConfig = if is_multi_population_value_type(value_agg.aggregation_type)
     {
-        let is_key_agg_on_metric = |c: &&AggregationConfig| {
-            c.metric == requirements.metric && is_key_agg_type(c.aggregation_type)
-        };
+        // Single pass: take the first window-valid key agg on this metric,
+        // but also remember the first window-*invalid* one seen (e.g. a
+        // Sliding DeltaSetAggregator) so the miss path below can report the
+        // specific reason instead of a generic "none found" (#588).
+        let mut invalid_window: Option<&AggregationConfig> = None;
         let ka = configs.values().find(|c| {
-            is_key_agg_on_metric(c) && key_agg_window_valid(c.aggregation_type, c.window_type)
+            if c.metric != requirements.metric || !is_key_agg_type(c.aggregation_type) {
+                return false;
+            }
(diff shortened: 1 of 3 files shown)


milindsrivastava1997:
Reviewed — build/tests pass, no correctness regressions found in the DeltaSetAggregator/Sliding fix (#588).

Two minor cleanups:
- `candidate_gen.rs:57` duplicates the DeltaSetAggregator+Sliding rejection inline instead of calling the new `key_agg_window_valid` predicate — could drift if the invariant is extended later.
- `capability_matching.rs:285` scans `configs.values()` twice on the miss path just to build a more detailed warn log.

PR #611, opened 2026-08-25
fix(query-engine): exact-window queries take a read lock, not write

### LINKED ISSUE #607: MutableEpoch::exact_query requires a write lock, serializing concurrent queries
`MutableEpoch::exact_query` (`asap-query-engine/src/stores/simple_map_store/common.rs:238`) takes `&mut self` because it lazily builds `window_to_ids: Option<HashMap<TimestampRange, Vec<u32>>>` on first access after any write, invalidating it on every `insert`. Callers (`per_key.rs:620`, `global.rs:568`, and legacy equivalents) must therefore take a **write lock** on the whole per-aggregation store shard for any exact-window query — even though the query itself doesn't mutate stored data, only an internal index.

By contrast:
- `query_precomputed_output` (the range/scan call) only takes a **read lock** (`per_key.rs:477`, `.read()`).
- `SealedEpoch::exact_query` (`common.rs:350`) already takes `&self` and does a binary search — sealed (immutable) epochs have no lock problem at all.

So the write-lock requirement is scoped to exactly one struct: the *current, still-being-written* `MutableEpoch`'s lazy index.

Effect: any code path that calls `query_precomputed_output_exact` serializes against all other readers/writers of that aggregation's current epoch for the query's duration — a real concurrency cost, not just latency. This matters more once #608 routes more traffic through the exact-query call (currently only Sliding-window instant queries use it).

## Proposed fix

Give `MutableEpoch::exact_query` a `&self` signature via interior locking scoped to just the index, e.g.:
- An atomic "dirty" flag (replacing the plain `window_to_ids = None` field write on insert — should stay ~zero-cost, an atomic store instead of a bare field write).
- The index itself behind its own small `RwLock<Option<HashMap<TimestampRange, Vec<u32>>>>` (or similar), rebuilt under that inner lock only when dirty.

This lets `per_key.rs`/`global.rs` (and their legacy equivalents) always take the outer per-shard lock as `.read()`, for both scan and exact queries. Insert's cost should remain effectively unchanged (still O(1) amortized, no HashMap work on the hot path).

## Scope

- `asap-query-engine/src/stores/simple_map_store/common.rs` (`MutableEpoch`)
- `asap-query-engine/src/stores/simple_map_store/per_key.rs` and `global.rs` (lock acquisition at the call sites)
- Legacy equivalents under `simple_map_store/legacy/` — check whether they're still live or dead code before touching.

## Testing

Needs both a correctness test (index still returns correct results across concurrent insert+query, dirty flag correctly forces rebuild) and, if feasible, a concurrency test demonstrating that two exact queries (or an exact query + a scan) no longer block each other.

Related: #581 (this removes a blocker for routing more of #581's fetch-mode unification through the exact-query call), #608.

milindsrivastava1997:
## Summary
- `MutableEpoch::exact_query` needed `&mut self` solely to lazily build/cache its `window_to_ids` offset index, forcing `per_key.rs`'s exact-query call site to take a full write lock on the per-aggregation shard even though the queried data itself isn't mutated — serializing exact-window queries against every other concurrent reader/writer of that shard.
- `window_to_ids` now lives behind its own `RwLock`, independent of the outer shard lock. `exact_query` takes `&self` and builds the index lazily under double-checked locking (read fast path; write + re-check slow path). `insert`/`remove_windows` still invalidate it at zero cost via `get_mut()` — no lock acquired, since `&mut self` already proves exclusive access there.
- `per_key.rs`'s call site drops from `.write()` to `.read()` — this is the actual concurrency fix (real `RwLock`-backed shard, the default `LockStrategy`).
- `global.rs` is updated to compile against the new signature (`get_mut` → `get`) but has no locking change — it uses a single `Mutex` for the whole store with no read/write split, so there's nothing to fix there.
- `legacy/` store implementations are untouched — confirmed dead code (unreachable via `LockStrategy`, no references anywhere outside `legacy/` itself).

## Testing
Test coverage was designed by a separate agent given only the problem description (not this fix), working from an isolated worktree checked out at the pre-fix commit, specifically to avoid biasing the tests toward this implementation's shape.

- 5 correctness tests added to the existing `run_contract_suite` (run against both `LockStrategy::PerKey` and `::Global`): stability across repeated calls, no shadowing of a later insert by an earlier cache build, miss-then-hit after insert, 900 interleaved insert/query rounds checked by value, and correctness across epoch rotation.
- 1 concurrency test (`test_exact_query_does_not_block_concurrent_range_queries_per_key`), written as a TDD red test asserting the desired *fixed* behavior: a concurrent range query's max latency should stay under 20% of a forced index-rebuild's duration. Pre-fix, this failed cleanly and consistently (reader latency ≈ exact-query duration, ratio ~1.00 across repeated runs) — a reader visibly blocked for the full write-lock hold. Post-fix, it passes.

Fixes #607.

## Test plan
- [x] `cargo test -p query_engine_rust --lib` — 576 passed, 0 failed
- [x] `cargo test -p query_engine_rust --lib store_correctness_tests` — `contract_per_key`, `contract_global`, and the new concurrency test all pass
- [x] `cargo fmt` / `cargo clippy` clean (pre-commit hooks passed)

### DIFF
--- a/asap-query-engine/src/stores/simple_map_store/INDEX_DESIGN.md
+++ b/asap-query-engine/src/stores/simple_map_store/INDEX_DESIGN.md
@@ -56,8 +56,8 @@ MutableEpoch {
     // Monotonic ingest fast path (Opt 3)
     last_window:     Option<TimestampRange>
 
-    // Lazy offset index (Opt 1 + 2): built on first exact_query, None after any insert
-    window_to_ids:   Option<HashMap<TimestampRange, Vec<u32>>>
+    // Lazy offset index (Opt 1 + 2): built on first exact_query, cleared on any insert
+    window_to_ids:   OnceLock<HashMap<TimestampRange, Vec<u32>>>
 
     // Epoch bounds for O(1) skip check (updated incrementally on insert)
     min_start:       Option<u64>
@@ -68,14 +68,15 @@ MutableEpoch {
 **Insert** (`O(1)` amortized):
 - Opt 3: if incoming window == `last_window`, skip `windows_set.insert` entirely
 - Three `Vec::push` calls — no secondary index maintenance
-- `window_to_ids = None` — single pointer-width write to invalidate the index
+- `window_to_ids.take()` — through `&mut self`, so no synchronization needed to invalidate the index
 
 **`seal()` → `SealedEpoch`** (`O(M log M)`, paid once at rotation):
 - Zips the three columns into tuples, sorts by `(TimestampRange, MetricID)`, moves `Arc`s without cloning
 
-**`exact_query(&mut self)`** (`O(M)` first call after a write, `O(m)` cached):
-- Opt 1 + 2: if `window_to_ids` is `None`, build it from `windows_col` in one pass storing `u32` offsets
-- Cache is valid until the next `insert`
+**`exact_query(&self)`** (`O(M)` first call after a write, `O(m)` cached):
+- Opt 1 + 2: `window_to_ids.get_or_init(...)` builds the index from `windows_col` in one pass storing `u32` offsets if not already built; otherwise returns the cached index directly
+- `OnceLock` gives build-at-most-once semantics under concurrent callers for free — no external locking needed to call this, and no poisoning risk if the build closure ever panicked
+- Cache is valid until the next `insert` (or `remove_windows`), which calls `window_to_ids.take()`
 
 **`range_query_into`** (`O(M)` mutable epoch):
 - Opt 5: hot loop iterates only `windows_col`; aggregate pointer only chased on match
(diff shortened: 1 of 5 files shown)


milindsrivastava1997:
## Code Review

Reviewed the core change (scoping `exact_query` to a read lock instead of write) across `common.rs`, `global.rs`, `per_key.rs`, and the new correctness tests. The locking logic is race-free and the exclusive/shared lock split between insert/remove_windows and exact_query is correct for both `PerKey` and `Global` stores. No correctness bugs in the main change.

Two minor findings on the new inner `window_to_ids` lock in `common.rs`:

1. **Poisoning cascade (line 185, `correctness`, plausible)**: Wrapping `window_to_ids` in its own `RwLock` introduces a new failure mode. If the index-build in `exact_query`'s write-lock slow path (lines 262-269) ever panics, only the inner `window_to_ids` lock gets poisoned — the outer per-shard lock stays healthy. But `insert()` (line 185) and `remove_windows()` (line 320) call `.get_mut().unwrap()` on it with no graceful handling, so every future write to that epoch panics indefinitely instead of surfacing as a `Result::Err`, the way outer-lock poisoning was handled pre-PR.

2. **Simplification (line 114, plausible)**: The manual double-checked-locking on `RwLock<Option<HashMap<...>>>` reimplements `std::sync::OnceLock` (stable since 1.70). `OnceLock::get_or_init` would replace the hand-rolled read-fast-path/write-slow-path/recheck logic (lines 251-277) with one call, and `OnceLock::take()` (needs `&mut self`, available at both invalidation sites) replaces the `get_mut().unwrap() = None` pattern at lines 185/320. It also sidesteps finding #1: a panicking `OnceLock` initializer leaves the cell uninitialized rather than poisoned, so index-build failures would self-heal instead of permanently panicking future writes.

zzylol:
lgtm

PR #627, opened 2026-08-26
feat(store): added store interface for batched exact-window queries

### LINKED ISSUE #609: Add a batched exact-window query API to the store
Once #608 routes Sliding-window range queries through `query_precomputed_output_exact` (one call per output step instead of one wide scan), a range query over N output steps issues N sequential exact-window store calls where it previously issued one scan call. Even after #607 removes the write-lock cost, that's still N separate lock-acquire-and-lookup round trips instead of one.

## Proposal

Add a batched variant, e.g. `query_precomputed_output_exact_batch(metric, aggregation_id, windows: &[TimestampRange]) -> Result<Vec<TimestampedBucketsMap>, _>` (exact shape TBD), that acquires the store shard lock once and resolves all requested exact windows against `MutableEpoch`/`SealedEpoch` in a single pass, instead of one lock cycle per window.

## Scope

- `asap-query-engine/src/stores/traits.rs` — new trait method.
- `asap-query-engine/src/stores/simple_map_store/{per_key,global}.rs` — implementation.
- `asap-query-engine/src/engines/simple_engine/mod.rs` — range pipeline's per-step Sliding fetch (from #608) becomes one batched call up front instead of a call inside the per-step loop.

## Sequencing

Pure perf follow-up, no correctness bug riding on it — #608's fix is correct with sequential exact calls today. Land after #607 and #608, whenever convenient.

Related: #581, #607, #608.

milindsrivastava1997:
Adds `Store::query_precomputed_output_exact_batch`: resolves N windows against one (metric, aggregation_id) shard in a single lock acquisition instead of N.

- `per_key.rs`: one read-lock for the whole batch.
- `global.rs`: one process-wide mutex for the whole batch.
- `scan_windows_via_exact`: builds its grid upfront, one batched call instead of a per-window loop — this loop fed range queries, instant tumbling queries, and keys queries alike after #616/#621.
- Tests: empty/unknown/invalid windows, sequential-call equivalence, read-count batching for ReadBased cleanup.

Closes #609.

### DIFF
--- a/asap-query-engine/src/stores/simple_map_store/global.rs
+++ b/asap-query-engine/src/stores/simple_map_store/global.rs
@@ -2,8 +2,8 @@ use crate::data_model::{
     AggregateCore, AggregationType, CleanupPolicy, PrecomputedOutput, StreamingConfig,
 };
 use crate::stores::simple_map_store::common::{
-    sort_buckets_chronologically, EpochID, InternTable, MetricBucketMap, MutableEpoch, SealedEpoch,
-    TimestampRange,
+    resolve_exact_windows, sort_buckets_chronologically, EpochID, InternTable, MetricBucketMap,
+    MutableEpoch, SealedEpoch, TimestampRange,
 };
 use crate::stores::{Store, StoreResult, TimestampedBucketsMap};
 use std::collections::{BTreeMap, HashMap, HashSet};
@@ -702,7 +702,26 @@ impl Store for SimpleMapStoreGlobal {
         let query_start_time = Instant::now();
         let store_key = aggregation_id;
 
+        // Measure lock acquisition time
+        #[cfg(feature = "lock_profiling")]
+        let lock_wait_start = Instant::now();
+
         let mut data = self.lock.lock().unwrap();
+
+        #[cfg(feature = "lock_profiling")]
+        {
+            let lock_wait_duration = lock_wait_start.elapsed();
+            info!(
+                "🔒 Batched exact query lock wait time: {:.2}ms (metric: {}, agg_id: {}, windows: {})",
+                lock_wait_duration.as_secs_f64() * 1000.0,
+                metric,
+                aggregation_id,
+                windows.len()
+            );
+        }
+
+        #[cfg(feature = "lock_profiling")]
+        let lock_hold_start = Instant::now();
 
         let per_key = match data.stores.get(&store_key) {
             Some(pk) => pk,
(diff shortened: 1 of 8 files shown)


asap-query-engine/src/stores/simple_map_store/per_key.rs
```diff
@@ -756,6 +756,96 @@ impl Store for SimpleMapStorePerKey {
         Ok(results)
     }
 
+    /// Batched exact-window lookup (#609): acquires the shard's read lock once for the
+    /// whole `windows` slice instead of once per window, resolving each window against
+    /// `current_epoch` / `sealed_epochs` in a single pass. Otherwise identical semantics
+    /// to calling `query_precomputed_output_exact` once per window and merging the
+    /// results (a window with no exact match simply contributes nothing).
+    fn query_precomputed_output_exact_batch(
```
milindsrivastava1997, line 764: The new batched exact-query methods drop the `lock_profiling` instrumentation (wait/hold time logging) present on the single-window `query_precomputed_output_exact`.

Building with `--features lock_profiling` to diagnose lock contention on range/instant queries: `query_precomputed_output_exact_batch` (here and in global.rs:692) never emits the lock wait/hold time logs that the single-window path emits, so exactly the code path most affected by the increased lock-hold-duration (see other comment) is invisible to that diagnostic tooling.

asap-query-engine/src/stores/simple_map_store/global.rs
```diff
@@ -685,6 +685,89 @@ impl Store for SimpleMapStoreGlobal {
         Ok(results)
     }
 
+    /// Batched exact-window lookup (#609): acquires the process-wide lock once for the
+    /// whole `windows` slice instead of once per window. Otherwise identical semantics to
+    /// calling `query_precomputed_output_exact` once per window and merging the results
+    /// (a window with no exact match simply contributes nothing).
+    fn query_precomputed_output_exact_batch(
```
milindsrivastava1997, line 692: `query_precomputed_output_exact_batch` is duplicated almost line-for-line between this file and per_key.rs (epoch lookup loop, read-count update loop, debug logging), adding ~90 more lines of near-identical logic that must be kept in sync by hand.

A future fix to the exact-match resolution order (current_epoch then sealed_epochs newest-to-oldest) or to the read-count bookkeeping could get applied to one copy (per_key.rs:801-831) and forgotten in the other (here, 722-753), silently reintroducing a bug in one of the two lock strategies — the contract test suite may not catch a strategy-specific regression.

Issue #581, opened 2026-08-23
Unify PromQL instant and range query fetch/merge paths


milindsrivastava1997:
PromQL instant queries and range queries use two independently implemented fetch/merge designs for the same underlying job, not one design expressed twice.

- Instant: one targeted fetch per `values_query` (exact-window if Sliding, range if Tumbling), merged once via `merge_precomputed_outputs` (fold over `merge_with`, with a `do_merge` short-circuit when only one bucket is in range).
- Range: one big contiguous fetch over the whole extended range, then a fresh `WindowMerger`/`NaiveMerger` built and fully re-merged from scratch per output step, per key. The trait's incremental `slide()` (add/remove) is never used.
- Key resolution is duplicated too: collect_all_results/collect_results_separate_keys/collect_results_same_aggregation (instant) vs the inline is_dual_population/get_keys()/fallback logic in execute_range_query_pipeline (range) independently reimplement "which output keys does this group serve." Already caused 3 separate bugs (#570, #582, #587) from the two paths drifting apart. Should collapse to one shared resolver both paths call.

Binary-expr instant and range queries each just run the corresponding plain-query pipeline once per arm and combine client-side, so this fork is duplicated across all four PromQL query paths, not just the two plain ones.

milindsrivastava1997:
## `is_exact_query` still hardcoded false for range queries — #580 not fully fixed

`finish_range_context` (`asap-query-engine/src/engines/simple_engine/promql.rs:585`) still unconditionally sets `extended_store_plan.values_query.is_exact_query = false`, regardless of the aggregation's actual `WindowType`. This is exactly problem #2 named in #580, which PR #582 closed — but #582's changes (keys_query expansion, same-timestamp bucket collision) never touched this line. The bug is still live on main.

Effect: a PromQL range query over a **Sliding**-window aggregation gets routed to the overlap-scan fetch (`query_precomputed_output`) instead of the exact-window fetch (`query_precomputed_output_exact`), and is then merged downstream as if it were Tumbling (`window_type = if is_exact_query {Sliding} else {Tumbling}`, `simple_engine/mod.rs:545`) — wrong values, silently, no error.

Instant queries don't have this bug: `create_store_query_plan` correctly derives `is_exact_query = window_type == WindowType::Sliding` (`simple_engine/mod.rs:414`). Only the range path hardcodes it away.

[...]

Issue #554, opened 2026-08-20
Sliding window execution: support multi-window merge/subtract in query engine


milindsrivastava1997:
## Problem

Sliding windows in the query engine today are **not** "tumbling with
configurable overlap" — they're a fundamentally narrower serving model. Per
`asap_types::capability_matching::window_compatible`
(`capability_matching.rs:63-77`):

```rust
match config.window_type {
    WindowType::Sliding => data_range_ms == window_ms,
    WindowType::Tumbling => data_range_ms.is_multiple_of(window_ms),
}
```

A sliding-window aggregation can only serve a query whose requested time
range matches the window's width **exactly** — no merge, no subtract, across
multiple stored overlapping windows. This is enforced live, on every query,
via `find_compatible_aggregation()` (called from both
`simple_engine/promql.rs:1062` and `simple_engine/sql.rs:503`). Tumbling
windows have had multi-bucket merge support all along
(`simple_engine/mod.rs`'s `do_merge` path); sliding never got the equivalent.

Concretely: `rate(http_requests_total[5m])` refreshed every 60s has
`data_range_ms = 300,000` but `window_size_ms = 60,000` (derived from
`t_repeat_ms`, unrelated to the query's own range). Under sliding this is
rejected outright — not wrong data, just "no compatible aggregation," always.
The only naturally-compatible case today is a spatial-only query with no
range literal at all, which is also the least interesting case to slide.

This blocks #555 (a manual config override to choose tumbling vs.
sliding per experiment) from being safely shippable — enabling it today
would produce plans that fail at query time for almost any real workload
query.

## Relationship to #405

#405 ("Optimization-based sketch/streaming config selection") already
documents `Sliding: W < range_a → Infeasible` as a **deliberate v1 modeling
assumption** in its own feasibility table and cost formulas — not an
unimplemented case. This issue is effectively a proposal to extend that
formulation (and the corresponding query-engine implementation) to support
`W < range_a` for sliding windows via merge/subtract, the same way tumbling
already works. #405 remains the design/tracking issue for automatic
optimizer-driven selection generally; this issue is scoped specifically to
making sliding-window *execution* correct, independent of who chooses to use
one (the optimizer, or the manual override in #555).

## What's already correct (not part of this issue)

The *ingestion* side already works and is tested:
`precompute_engine/window_manager.rs`'s `WindowManager` correctly produces a
dense, Unix-epoch-aligned grid of overlapping window buckets, one every
[...]

Issue #23, opened 2026-03-04
Re-architect code to use Redis for control messages


milindsrivastava1997:
Atleast 2 re-architecturing changes come to mind:

- remote_monitor should not be orchestrating PrometheusClient. It is only doing right now so that it knows when to start and stop recording resource usage. 
- PrometheusClient should not be orchestrating profilers. It is only doing so right now so that it knows when to start and stop profiling

Both these usecases (and others) should be solved by the "master orchestrator" (currently, experiment_run_e2e.py) sending control messages to various components

PR #629, opened 2026-08-26
feat(query-engine): computed top-k in range queries + stage-E equivalence tests (#581 stage E prep)

milindsrivastava1997:
## Summary
- Range queries previously had zero top-k support at all (`execute_range_query_pipeline` had no limiting/formatting params). Adds step-major top-k ranking/truncation (`apply_range_topk`) — ranks and truncates independently at each output timestamp, since a range query's surviving key set can differ step to step, unlike instant's single-value-per-group case. Wired through all three PromQL range call sites.
- Adds an instant/range equivalence test matrix ahead of stage E's full pipeline collapse: 4 window shapes (Tumbling ratio 1/2/3, one genuinely-overlapping Sliding case) × {no keys, SetAggregator keys, DeltaSetAggregator keys} × {instant, range}, plus 3 dedicated tests for the new topk plumbing (step-major re-ranking across steps, single-step-vs-instant equivalence, non-interference with #595's self-keyed sketch expansion).
- Two corrections to the originally-scoped test matrix, discovered while writing it (documented in the new test file's header): a Sliding window's `lookback_ms` must equal `window_size_ms` (active `assert!` in `execute_range_query_pipeline`, tracked separately as #554), so "Sliding at ratio 2/3" isn't constructible; and `SetAggregator`/`DeltaSetAggregator` are keys-side types whose window_type is independent of the paired value aggregation's, so no cells needed excluding.
- Rebased onto main post-#627/#628 merge; no functional overlap (different regions of `mod.rs`/`promql.rs`).

Note: "computed top-k" over an arbitrary non-self-keyed expression (e.g. `topk(5, rate(foo[5m]))`) turns out to be unsupported everywhere in this codebase today, not just in range (`SumAccumulator::query` hard-errors on `Statistic::Topk`). Out of scope here, left as-is.

## Test plan
- [x] `cargo test -p query_engine_rust --lib`: 520/520 passing
- [x] `cargo clippy -p query_engine_rust --lib --tests -- -D warnings`: clean
- [x] Roborev review (job #23): 1 Low finding fixed (dead-code `retain`), 1 Low finding (tie-break nondeterminism) left as-is per reviewer's own note — pre-existing instant-side behavior, not introduced here

### DIFF
--- a/asap-query-engine/src/engines/simple_engine/mod.rs
+++ b/asap-query-engine/src/engines/simple_engine/mod.rs
@@ -1921,12 +1921,14 @@ impl SimpleEngine {
 
                 results.retain(|key, _| kept_timestamps_by_key.contains_key(key));
                 for elem in results.values_mut() {
+                    // `keep` is built only from timestamps that already
+                    // appear in this same element's `samples` (see the
+                    // `by_timestamp` loop above), and is non-empty for every
+                    // key that survives the `retain` just above -- so this
+                    // filter can never leave `elem.samples` empty.
                     let keep = &kept_timestamps_by_key[&elem.labels];
                     elem.samples.retain(|s| keep.contains(&s.timestamp));
                 }
-                // A group that made no timestamp's top-k has no samples left
-                // -- drop it entirely rather than emitting an empty series.
-                results.retain(|_, elem| !elem.samples.is_empty());
             }
         }
 
--- a/asap-query-engine/src/engines/simple_engine/mod.rs
+++ b/asap-query-engine/src/engines/simple_engine/mod.rs
@@ -1500,10 +1500,20 @@ impl SimpleEngine {
         }
     }
 
-    /// Execute the range query pipeline
+    /// Execute the range query pipeline.
+    ///
+    /// `enable_topk_limiting`/`enable_topk_formatting` mirror
+    /// `execute_query_pipeline`'s flags of the same name (see that method's
+    /// doc comment) -- both no-ops unless
+    /// `context.base.metadata.statistic_to_compute == Statistic::Topk`. The
+    /// actual ranking/truncation is delegated to `apply_range_topk` below;
+    /// see its doc comment for why range's version can't just reuse
+    /// instant's `format_final_results` truncate-once shape.
     fn execute_range_query_pipeline(
         &self,
         context: &RangeQueryExecutionContext,
+        enable_topk_limiting: bool,
+        enable_topk_formatting: bool,
     ) -> Result<Vec<crate::engines::query_result::RangeVectorElement>, String> {
         use crate::engines::query_result::RangeVectorElement;
         use crate::engines::window_merger::create_window_merger;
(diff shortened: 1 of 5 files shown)


asap-query-engine/src/engines/simple_engine/promql.rs
```diff
@@ -745,8 +748,13 @@ impl SimpleEngine {
         if lhs_labels != rhs_labels {
             return None;
         }
-        let lhs_results = self.execute_range_query_pipeline(&lhs_ctx).ok()?;
-        let rhs_results = self.execute_range_query_pipeline(&rhs_ctx).ok()?;
+        // (true, true): self-gated, same rationale as the scalar-arm call above.
+        let lhs_results = self
+            .execute_range_query_pipeline(&lhs_ctx, true, true)
```
milindsrivastava1997, line 753: **Correctness (CONFIRMED):** Vector-vector range binary ops over two `topk(...)` arms silently return an empty matrix. `execute_range_query_pipeline(..., true, true)` runs `apply_range_topk` with `enable_topk_formatting=true`, which prepends each arm's own metric name into `elem.labels` (mod.rs:1935-1943). `rhs_map` is keyed by `elem.labels` (line 767) and `lhs_elem.labels` is looked up directly (line 772) — since the label vectors now start with different metric-name strings, the join never matches and the result is an empty `Matrix` instead of combined per-group sums. Reproduced: `topk(2, metric_a) + metric_b` as a range query. This is a regression vs. pre-PR behavior, where `execute_range_query_pipeline` never mutated labels for range queries.

asap-query-engine/src/engines/simple_engine/mod.rs
```diff
@@ -1829,8 +1839,112 @@ impl SimpleEngine {
...
+
+                // Step-major ranking: group every group's samples by
+                // timestamp first, so each timestamp's top-k decision sees
+                // every group's value at that timestamp.
+                let mut by_timestamp: HashMap<u64, Vec<(KeyByLabelValues, f64)>> = HashMap::new();
+                for elem in results.values() {
+                    for sample in &elem.samples {
+                        by_timestamp
+                            .entry(sample.timestamp)
+                            .or_default()
```
milindsrivastava1997, line 1927: **Correctness (CONFIRMED):** Step-major top-k ranking iterates `results` (a `HashMap`) without a tiebreak before the stable sort. With two groups tied at the k-th value (e.g. two groups both at 50.0, k=2), whichever iterates first (HashMap order is unordered across runs/hasher seeds) keeps its sample at that timestamp and the other is dropped — the surviving key set for a tied topk query is non-deterministic run to run, not just its order.

asap-query-engine/src/engines/simple_engine/mod.rs
```diff
@@ -1829,8 +1839,112 @@ impl SimpleEngine {
...
+        if enable_topk_limiting {
+            if let Some(k) = query_kwargs.get("k").and_then(|s| s.parse::<usize>().ok()) {
+                use std::collections::HashSet;
+
+                // Step-major ranking: group every group's samples by
+                // timestamp first, so each timestamp's top-k decision sees
+                // every group's value at that timestamp.
+                let mut by_timestamp: HashMap<u64, Vec<(KeyByLabelValues, f64)>> = HashMap::new();
+                for elem in results.values() {
+                    for sample in &elem.samples {
```
milindsrivastava1997, line 1924: **Efficiency (PLAUSIBLE):** `by_timestamp.entry(...).push((elem.labels.clone(), sample.value))` clones the full label `Vec<String>` once per (group, timestamp) sample instead of once per group — for G groups over T steps that's G*T clones where G would suffice if keyed by a lightweight index instead.

asap-query-engine/src/engines/simple_engine/mod.rs
```diff
@@ -1829,8 +1839,112 @@ impl SimpleEngine {
...
+                    // `by_timestamp` loop above), and is non-empty for every
+                    // key that survives the `retain` just above -- so this
+                    // filter can never leave `elem.samples` empty.
+                    let keep = &kept_timestamps_by_key[&elem.labels];
+                    elem.samples.retain(|s| keep.contains(&s.timestamp));
+                }
+            }
+        }
+
+        if enable_topk_formatting {
```
milindsrivastava1997, line 1967: **Correctness (PLAUSIBLE):** `apply_range_topk` skips ranking/sorting when `enable_topk_limiting` is false even if `enable_topk_formatting` is true, unlike the instant path which always sorts Topk results regardless of the limiting flag. Currently unreachable (all call sites pass `(true, true)` together), but the flags are threaded as independently documented parameters mirroring `execute_query_pipeline`'s contract — a future caller passing `(false, true)` would silently diverge from the analogous instant-path behavior.

milindsrivastava1997 (commented):
Automated review findings.

asap-query-engine/src/engines/simple_engine/promql.rs
milindsrivastava1997: Tracked in https://github.com/ProjectASAP/ASAPQuery/issues/631

milindsrivastava1997: Tracked in https://github.com/ProjectASAP/ASAPQuery/issues/631

milindsrivastava1997: Doesn't reproduce as described (pre-existing __name__ label-name mismatch rejects it before the join runs); real repro + Finding-1 fix tracked in #631, not this PR.

asap-query-engine/src/engines/simple_engine/mod.rs
milindsrivastava1997: Fixed in 8551d2f: added a label-values tiebreak to the sort, confirmed via a flaky RED test (4/5 pass pre-fix) now stable across 20+ runs.

asap-query-engine/src/engines/simple_engine/mod.rs
milindsrivastava1997: Fixed in 8551d2f: indexed each group once instead of cloning its label vector per (group, timestamp) sample.

asap-query-engine/src/engines/simple_engine/mod.rs
milindsrivastava1997: Documented in 8551d2f: range's HashMap-backed results has no meaningful outer order, so this stays a no-op today rather than dead ranking logic.

Issue #594, opened 2026-08-24
Add differential e2e testing against real Prometheus using promql-compliance-tester


milindsrivastava1997:
## Problem

We have no differential/e2e test harness comparing ASAPQuery's PromQL output against real Prometheus. A large fraction of recent open bugs are exactly the class this would catch automatically: instant-vs-range query divergence (#589, #584, #583, #581, #587), missing counter-reset correction (#476), dropped outer aggregation (#496), sum/count semantics confusion (#505, #503). These keep getting found by hand, one at a time.

## Proposal

Adopt [`prometheus/compliance/promql`](https://github.com/prometheus/compliance/tree/main/promql) (the PromLabs `promql-compliance-tester`) as a differential test harness: it runs arbitrary PromQL query strings against a reference target and a test target and diffs the results, with per-test-case config for float tolerance (`AdjustValueTolerance{Fraction, Margin}`) and label-drop tweaks — which we need anyway since sketch-backed paths are approximate by design.

Two gaps to close before it's usable for us:

**1. It only tests range queries.** `Comparer.Compare` (in `comparer.go`) calls `QueryRange` exclusively; the `PromAPI.Query` (instant) method is declared in the interface but never invoked anywhere in the module. Since several of our worst bugs are specifically instant-vs-range divergence, a tool that never exercises the instant path can't catch them. We need to fork/patch `Comparer.Compare` to also run and diff an instant query per test case (~15-20 lines, mirrors the existing range logic).

**2. Its documented data-loading story doesn't fit us.** The upstream pattern is: run a real Prometheus scraping a demo exporter, mirror that scrape via `remote_write` into the target system, and wait ~1 hour for enough data to accumulate. That's non-deterministic and slow, and upstream itself lists "automate data loading" as an open TODO.

Instead: **write our own data seeder that pushes hand-authored samples directly via remote write** to both targets — build a `prompb.WriteRequest` with explicit timestamps/values/labels, snappy-compress it, POST to `/api/v1/write` on both a real Prometheus (started with `--web.enable-remote-write-receiver`) and ASAPQuery's existing remote-write endpoint. Same code path hits both targets, so there's no risk of the two ingestion mechanisms behaving differently and producing false diffs. Fully deterministic — we choose the exact data, so expected query results are computable by hand for regression cases.

## Plan

- [ ] Write the remote-write seeder (Go, `prompb` + snappy) — pushes a fixed dataset to two `query_url`s
[...]

Issue #590, opened 2026-08-24
Test suite ideas for instant/range PromQL query paths


milindsrivastava1997:
Test suite ideas for the instant-query/range-query pipelines in `asap-query-engine/src/engines/simple_engine/{mod,promql}.rs`, scoped alongside #589 (catalog of instant/range inconsistencies) and #584/#583/#581/#587.

## Differential oracle test
For each `(aggregation_type x window_type x single/dual-population x keyed/self-keyed)` combination, run an instant query at `t` and a range query with `start=end=t`, assert identical output sets. This is the generic test that would have caught #584 without needing a one-off regression case per bug.

## N-instants-equal-one-range property test
For a stable key set over `[start, end]`, assert `range(start, end, step)` equals `{instant(t) for t in steps}`. Then specifically break the "stable key set" assumption — a label present only in the first half of the range, and one present only in the second half — to pin down #583: assert absence in the un-populated steps, not just a wrong value.

## Keys-window-type coverage crossed with range
`create_keys_query_params`'s `SetAggregator` and `DeltaSetAggregator` branches are likely only exercised by instant-query tests today. Add the same matrix under range queries.

## Sliding-exact-window vs single-step-range parity
A range query covering exactly one Sliding window (step == window size) must equal the instant exact-query result bit-for-bit. Targets the fetch-strategy divergence noted in #589 item 4.

## Binary-expr mirrors of every plain-query regression
Top-k self-keyed metric over a range inside a binary expr; dual-population metric with a changing key set combined with itself over a range. The binary-expr path is a separate fork (see #589 item 6), so every plain-query regression test needs a binary-expr counterpart.

## Error-path parity
Missing key mid-merge should either fail both instant and range the same way, or the divergence should be an explicit, named, tested behavior (see #589 item 7).

## Boundary/validation tests
- `step` not a multiple of `tumbling_window_ms` (validate_range_query_params)
- `start == end`
- single-bucket range
- partial data mid-window (should still produce a sample)
- zero data anywhere in window (should skip that step, not error)
- missing keys for a group (should error, per #582)

## Model-based test using existing simulation helpers
`simulate_sliding_window` / `simulate_sliding_window_with_alignment` (mod.rs:1728+) are already isolated reference oracles used for hand-picked cases — extend them to cover key expansion and use as the oracle for a broader property-based test rather than only fixed cases.

Related: #589, #584, #583, #581, #587.

PR #630, opened 2026-08-26
test(query-engine): add differential PromQL testing infra (#590, #594)

milindsrivastava1997:
## Summary

Parallel test-infrastructure work for #590 and #594, run as 5 scoped agents against genuinely uncovered gaps (the existing suite already covers a lot of instant/range regression territory — this fills what was left).

**#590 — Rust test suite for instant/range PromQL paths** (`asap-query-engine`)
- `range_multistep_instant_equivalence_tests.rs` — the generic oracle the issue called out as missing: a multi-step range query vs. N independent instant queries, one per step, across Tumbling/Sliding × single/dual-population. All pass (stable-key-set happy path).
- `range_query_validation_tests.rs` + inline tests in `simple_engine/mod.rs` — boundary coverage for `validate_range_query_params` (start>=end, step==0, step-not-multiple-of-window, start==end), plus an instant/range skip-behavior parity check for orphan groups.
- `sliding_window_keyed_oracle_tests.rs` + new `simulate_sliding_window_keyed` reference oracle — a property sweep (21 scenarios) over keyed sliding-window merges.

**Bug found:** the sliding-window sweep surfaced a real bug in `execute_range_query_pipeline` — for a step with `current_time < window_size_ms`, `window_start = current_time.saturating_sub(window_size_ms)` clamps to `0`, aliasing onto the same store bucket as the legitimate step at `current_time == window_size_ms`. The too-early step gets a phantom duplicate of the later step's value instead of correctly having no sample. Pinned as a minimal, `#[ignore]`d regression test (not fixed here — out of scope for this PR, tracked separately).

**#594 — Go differential e2e harness** (`promql-compliance/`)
- `seeder/` — remote-write dataset seeder (prompb + snappy) pushing a fixed, hand-authored dataset to two `query_url`s, so both a reference Prometheus and ASAPQuery ingest through the identical code path. Includes a deliberately gapped series to probe instant-vs-range divergence.
- `harness/` — forked `prometheus/compliance/promql`, with a surgical patch to `Comparer.Compare` so it also runs and diffs instant queries (previously range-only), plus `config.yaml` seeded with 4 regression cases tagged to #583/#584/#589.

No live Prometheus/ASAPQuery run yet — this PR is code-only for the Go harness, per plan. CI wiring is deliberately deferred.

## Test plan
- [x] `cargo fmt`, `cargo check`, `cargo clippy`, `cargo test` all pass (full pre-commit suite, run on the merged result)
- [x] `go build ./...` and `go vet ./...` pass for both `promql-compliance/seeder` and `promql-compliance/harness`
- [ ] File a follow-up issue for the phantom-window sliding-range bug
- [ ] Wire the Go harness into CI (separate PR, needs live Prometheus/ASAPQuery target decision)


milindsrivastava1997:
## Handoff

Built via 5 parallel agents against genuine gaps in existing coverage (a lot of instant/range regression territory was already tested — this fills what wasn't).

### Done, in this PR
- **#590 (Rust)**: generic multi-step instant≡range oracle (`range_multistep_instant_equivalence_tests.rs`), `validate_range_query_params` boundary tests (`range_query_validation_tests.rs`), keyed sliding-window property oracle (`sliding_window_keyed_oracle_tests.rs` + new `simulate_sliding_window_keyed` reference fn). All green.
- **#594 (Go, new `promql-compliance/`)**: `seeder/` (remote-write dataset pusher, prompb+snappy) and `harness/` (forked `compliance/promql`, patched to diff instant queries alongside range queries, 4 seed cases tagged to #583/#584/#589). Code-only — not run against a live Prometheus/ASAPQuery yet.

### Found along the way
- **#632**: the keyed sliding-window oracle caught a real bug — Sliding-window range queries fabricate a phantom value for steps before a full window's history exists (`current_time.saturating_sub(window_size_ms)` aliases to the same bucket as a later, legitimate step). Pinned as an `#[ignore]`d regression in this PR, not fixed here.

### TBD / not in this PR
1. **Fix #632.** Not attempted — found by the tests this PR adds, scoped as its own fix.
2. **Live end-to-end run of the Go harness** — needs a real Prometheus (`--web.enable-remote-write-receiver`) and ASAPQuery instance to point `config.yaml` at; currently placeholder URLs (`localhost:9090`/`9091`).
3. **CI wiring for the Go harness** — last checkbox in #594, deliberately deferred (needs a decision on where/how it runs).
4. **#590/#594 issue bodies** still have their own checklists — worth revisiting once #632 is fixed and the harness has a live run, to decide what's left open vs. closeable.

### Review pointers
- The bug repro is the most interesting part to check first: `asap-query-engine/src/tests/sliding_window_keyed_oracle_tests.rs:305`.
- Go modules each have their own `go.mod` under `promql-compliance/seeder/` and `promql-compliance/harness/` — no root Go module, nothing wired into the Rust build.

milindsrivastava1997:
Design / implementation plan for the differential query runner:

- YAML fixtures define datasets as relative samples; one canonical remote-write payload is sent to both Prometheus and ASAPQuery.
- YAML query suites define explicit instant timestamps and range windows per query.
- Comparisons are exact by default. Tolerance is explicit, visible in config and JSON output, and can be set globally with per-query overrides.
- Each query checks Prometheus vs ASAPQuery for range and instant results, plus range-at-t vs instant-at-t within each target.
- Runs use an isolated Docker Compose project and fresh volumes by default, with generated configs and machine-readable JSON reports.
- The Go runner is the focused workflow; the existing copied compliance harness remains secondary until it is migrated or removed.

This is the shape being implemented in the PR 630 worktree, with tests and CI wiring following the vertical slices.

PR #643, opened 2026-08-27
feat(planner): add manual windowing override

### LINKED ISSUE #555: Manual windowing override (tumbling/sliding) in config.yaml
## Problem

`asap-planner-rs`'s live planner is hardcoded to always emit tumbling
windows. `should_use_sliding_window()` (`planner/window.rs:11-14`)
unconditionally returns `false`

`WindowType::Sliding` already exists in `asap_types::enums` and is exercised
by the offline optimizer/cost-model code (`optimizer/candidate_gen.rs`,
`cost_model.rs`), but never reachable from the live planner paths that
`asap-tools/experiments` actually drives.

## What this issue covers

Add a manual `windowing` override to `config.yaml`, threaded through to the
planner, so an experiment can explicitly choose tumbling vs. sliding windows
(and a slide granularity) instead of always getting tumbling:

```yaml
windowing:
  type: "tumbling"   # or "sliding"
  slide_divisor: 4    # only meaningful when type == "sliding";
                       # slide_interval_ms = window_size_ms / slide_divisor
```

- **Global** override (not per-query-group) for now — matches
  `sketch_parameters`'s existing global-only shape. Per-*aggregation-id*
  control is the ideal case but out of scope here (tracked separately).
- **PromQL + SQL** paths: `ControllerConfig`/`SingleQueryProcessor` →
  `set_window_parameters()`, and `SQLControllerConfig`/
  `SQLSingleQueryProcessor` → `compute_sql_window()`. ElasticDSL excluded —
  nothing in `asap-tools/experiments` generates its controller config today.
- `slide_divisor` must evenly divide each query's computed `window_size_ms`
  (which varies per query — it's derived from that query's own `t_repeat_ms`
  /`step_ms`, not a global constant) — error loudly, never round.
- **Fail-fast**: if any one query in the workload can't legally take a
  sliding window under the chosen divisor, plan generation aborts entirely
  rather than silently punting that query to the raw backend. This is a
  deliberate experiment setting, not an automatic optimizer decision — a
  silent fallback would quietly corrupt the cost/latency comparison the
  experiment exists to make.

## Blocked by #554

Shipping this override before sliding-window queries can actually be served
correctly (beyond the narrow single-window-exact-match case) means
`windowing.type: sliding` would produce plans that fail at query time for
almost any real workload query. See #554 for why, and for the
execution-side work this depends on.

## Definition of done

- Config plumbing + validation + fail-fast behavior, both PromQL and SQL
  paths.
- Unit tests for the plumbing itself (config parses; divisor-validation
  errors correctly; fail-fast behavior) — these don't depend on #554
  and can be written now.
- **Not required here**: an end-to-end "sliding query actually returns
  correct data" test. That belongs to #554, since it's the thing that
  proves *execution* works, not *candidate generation* — this issue is
  scoped to generating the candidates/config, not executing them.

## Related

- Complementary to #405 (optimization-based automatic config selection) —
  this is the manual stand-in for what #405 would eventually choose
  automatically, not a duplicate of it.
- Full grilling decisions log (rationale for every choice above) attached as
  a comment on this issue.

milindsrivastava1997:
## Summary
- add a global `windowing` override for PromQL and SQL planner inputs
- support explicit tumbling/sliding candidates with per-query divisor validation
- fail fast with aggregated validation errors and preserve query-log defaults
- thread the option through experiment config generation and document it

## Testing
- 41 PromQL integration tests
- 35 SQL integration tests
- repository pre-commit hooks

Execution-side behavior remains out of scope; this PR only controls planner candidate generation.

Closes #555

### DIFF
--- a/asap-planner-rs/src/config/input.rs
+++ b/asap-planner-rs/src/config/input.rs
@@ -10,6 +10,7 @@ use tracing::warn;
 #[serde(deny_unknown_fields)]
 pub struct ControllerConfig {
     pub query_groups: Vec<QueryGroup>,
+    pub windowing: Option<WindowingConfig>,
     pub sketch_parameters: Option<SketchParameterOverrides>,
     pub aggregate_cleanup: Option<AggregateCleanupConfig>,
     /// Optional hint: per-metric label sets used as a fallback when Prometheus
@@ -93,6 +94,21 @@ pub struct AggregateCleanupConfig {
     pub policy: Option<CleanupPolicy>,
 }
 
+#[derive(Debug, Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct WindowingConfig {
+    #[serde(rename = "type")]
+    pub window_type: WindowingType,
+    pub slide_divisor: Option<u64>,
+}
+
+#[derive(Debug, Clone, Copy, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum WindowingType {
+    Tumbling,
+    Sliding,
+}
+
 #[derive(Debug, Clone, Deserialize, Default)]
 pub struct SketchParameterOverrides {
     #[serde(rename = "CountMinSketch")]
@@ -142,6 +158,7 @@ pub struct HllParams {
 pub struct SQLControllerConfig {
     pub query_groups: Vec<SQLQueryGroup>,
     pub tables: Vec<TableDefinition>,
+    pub windowing: Option<WindowingConfig>,
     pub sketch_parameters: Option<SketchParameterOverrides>,
     pub aggregate_cleanup: Option<AggregateCleanupConfig>,
 }
--- a/asap-planner-rs/src/config/input.rs
+++ b/asap-planner-rs/src/config/input.rs
@@ -102,6 +102,24 @@ pub struct WindowingConfig {
     pub slide_divisor: Option<u64>,
 }
 
+impl WindowingConfig {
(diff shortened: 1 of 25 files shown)


milindsrivastava1997:
One real bug, three consistency/design gaps, one minor inefficiency.

**Real bug** (`asap-planner-rs/src/planner/window.rs:93`): `apply_windowing_override` checks `window_size_ms % slide_divisor == 0` but never checks `data_range_ms % window_size_ms == 0`, which `cleanup.rs`'s Sliding branch requires downstream. A realistic config (`rate(x[5m])` at `repetition_delay_ms: 45000`) passes the override check but fails in `cleanup.rs` with a differently-worded error that doesn't match the `"window_size_ms ("` prefix the generators use to aggregate errors — so it aborts on the first bad query instead of collecting all of them, breaking the PR's own "report all invalid queries" feature. Both new test fixtures happen to avoid this case entirely.

Two related design issues in the same aggregation mechanism: the string-prefix match is duplicated across `promql/generator.rs:109` and `sql/generator.rs:102` with no shared constant, and binary-arm queries (`promql/generator.rs:184`) bypass the aggregation `Vec` entirely via `?`.

Plus: `SQLController::from_file_with_discovery` (`sql/controller.rs:40`) skips `windowing.validate()` that every other constructor calls, and `windowing.validate()` runs redundantly up to 4x per plan generation (minor, not a correctness issue).

PR #657, opened 2026-09-01
fix(precompute): correct counter resets in rate and increase

### LINKED ISSUE #476: rate/increase functions do no counter-reset correction

milindsrivastava1997:
## Summary

- Correct PromQL `rate()`/`increase()` counter-reset handling for single and keyed accumulators.
- Track reset corrections through explicit reset events and preserve them across merges and serialization.
- Use the explicit-event-only `INC6` binary format.
- Remove V2/V5 compatibility branches and opaque reset metadata; old serialized payloads are intentionally rejected.
- Keep legacy `asap-summary-ingest` Arroyo templates out of scope.

## Cutover

This PR intentionally breaks compatibility with existing increase-accumulator JSON and binary payloads. Existing precomputes must be discarded or regenerated before deploying the new format. The active Arroyo adapter also requires the current reset-aware fields.

## Validation

- `cargo test -p query_engine_rust` — 582 tests passed.
- `cargo clippy -p query_engine_rust --lib --tests -- -D warnings` passed.
- Commit hooks passed formatting, check, clippy, and tests.

Fixes #476

### DIFF
--- a/asap-query-engine/src/precompute_operators/increase_accumulator.rs
+++ b/asap-query-engine/src/precompute_operators/increase_accumulator.rs
@@ -10,11 +10,18 @@ use promql_utilities::query_logics::enums::Statistic;
 
 pub(crate) const INCREASE_BINARY_FORMAT_MAGIC_V2: [u8; 4] = *b"INC2";
 pub(crate) const INCREASE_BINARY_FORMAT_MAGIC_V3: [u8; 4] = *b"INC3";
+pub(crate) const INCREASE_BINARY_FORMAT_MAGIC_V4: [u8; 4] = *b"INC4";
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct CounterResetEvent {
     pub timestamp: i64,
     pub adjustment: f64,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct OpaqueResetRange {
+    pub starting_timestamp: i64,
+    pub last_seen_timestamp: i64,
 }
 
 /// Accumulator for tracking increases in counter metrics
@@ -30,6 +37,8 @@ pub struct IncreaseAccumulator {
     pub counter_reset_adjustment: f64,
     #[serde(default)]
     pub counter_reset_events: Vec<CounterResetEvent>,
+    #[serde(default)]
+    pub opaque_reset_ranges: Vec<OpaqueResetRange>,
 }
 
 impl IncreaseAccumulator {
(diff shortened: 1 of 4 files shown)


milindsrivastava1997:
Compared against #492 (older PR fixing the same issue, #476): both use the same reset-detection/correction algorithm and equivalent merge semantics. The one functional gap here is Prometheus boundary extrapolation (`extrapolatedRate`) — #492 implements it, this PR doesn't. Extrapolation covers the small gap between the window's actual start/end and the nearest real sample on each edge; without it, `rate()`/`increase()` here will be measured only between the observed samples, so results will drift slightly from real Prometheus as scrape interval grows relative to the query window. Reset-loss correctness itself is fixed correctly in both.

